PR #45, opened 2015-09-11
Fix on the way message extension reads, so that if the message is too…

briangoncalves:
… long, it reads on small chunks (had issues when message was too long and was not reading correctly, truncating the result).

### DIFF
--- a/src/NServiceBus.AmazonSQS/Extensions/SqsMessageExtensions.cs
+++ b/src/NServiceBus.AmazonSQS/Extensions/SqsMessageExtensions.cs
@@ -1,6 +1,7 @@
 ﻿namespace NServiceBus.AmazonSQS
 {
 	using System;
+	using System.IO;
 	using Amazon.SQS.Model;
 	using Amazon.S3;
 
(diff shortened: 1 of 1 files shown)


ahofman:
Thanks Brian! Please forgive the lateness of my reply, my wife has just had a baby so we've been a little busy here :)

A few questions:

How big did messages have to be before you started seeing failures? 

Any particular reason you chose 8k for the chunk size?

briangoncalves:
I don’t know the minimum size for it to fail, but I had messages with a list of records that was getting about 400kb, and those was failing.

I noticed at the WriteResponseStreamToFileAsync method of the GetObjectResponse file, which is part of Amazon.S3 (AWSSDK), that it reads the buffer in 8k chunks, that’s why I followed the same rule, in case this is a limitation at Amazon side or Amazon SDK.

Brian Goncalves
Consultant, Information Technology
GCM Grosvenor
T +1-646-362-3613
M +1-917-216-6543
bgoncalves@gcmlp.com<mailto:bgoncalves@gcmlp.com>

ahofman:
Great, thanks @briangoncalves . I'm awaiting some feedback from some other users of the plugin to see if they have come across this issue yet.

Issue #50, opened 2016-08-03
Conflict when creating new buckets when in server farm environment


chrissimon-au:
When multiple servers attempt to create the same s3 buckets, an exception is thrown

```
Exception from CreateQueueIfNecessary.
Amazon.S3.AmazonS3Exception: A conflicting conditional operation is currently in progress against this resource. Please try again.
```

This causes the process to fail on launch.  Restart (manual or auto) resolves the problem as the buckets are now created and the risk of conflicting conditional operation being currently in progress goes down, but it would be nice if the plugin included auto-retry handling of this exception.  Not sure if the same thing can happen with queue creation.

Low priority for us as is easily work-aroundable, just recording the issue for completeness.

PR #53, opened 2017-02-12
bug 409 conflict on lifecyclerule

chrissimon-au:
**Note:** Please don't merge immediately - PR created for review, whilst proceeding with live testing.  Will advise when ready to merge.

Written to address #50, but looking at the history may also address #21.

As per the error response message from amazon, retry the lifecycle rule put operation in case of a conflict response.

* Get the lifecycle rule first to avoid the put operation if not needed (likely to be the case if there was a conflict due to multiple servers starting at the same time, or if the rule has been previously created)
* In case of conflict, retry up to 5 times, with linear backoff wait between retries (250ms * attemptCount)
* Throw original exception if unable to successfully put the rule after 5 attempts.

**Notes:**

* Have not been able to execute integration and acceptance tests in my local environment.  Hoping that the creation of this PR will trigger some automated test runs.  If they don't, let me know and I'll do a bit more to setup local environment and ensure they can execute before you merge.
* Included the Moq library in the NServiceBus.AmazonSQS.Tests project to facilitate mocking a 409 Conflict response for unit testing

### DIFF
--- a/src/NServiceBus.AmazonSQS.Tests/NServiceBus.AmazonSQS.Tests.csproj
+++ b/src/NServiceBus.AmazonSQS.Tests/NServiceBus.AmazonSQS.Tests.csproj
@@ -33,6 +33,14 @@
     <Reference Include="AWSSDK">
       <HintPath>..\packages\AWSSDK.2.3.15.0\lib\net45\AWSSDK.dll</HintPath>
     </Reference>
+    <Reference Include="Castle.Core, Version=3.3.0.0, Culture=neutral, PublicKeyToken=407dd0808d44fbdc, processorArchitecture=MSIL">
+      <HintPath>..\packages\Castle.Core.3.3.3\lib\net45\Castle.Core.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="Moq, Version=4.5.30.0, Culture=neutral, PublicKeyToken=69f491c39445e920, processorArchitecture=MSIL">
+      <HintPath>..\packages\Moq.4.5.30\lib\net45\Moq.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
     <Reference Include="NServiceBus.Core, Version=5.0.0.0, Culture=neutral, PublicKeyToken=9fc386479f8a226c, processorArchitecture=MSIL">
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\packages\NServiceBus.5.1.3\lib\net45\NServiceBus.Core.dll</HintPath>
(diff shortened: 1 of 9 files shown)


ahofman:
Hi @chrissimon-au , apologies for my lateness in replying! I have some time up my sleeve in the next few weeks so I will review this and feed back shortly. Thanks

ahofman:
In general I agree that it makes good sense to retry on a 409 response (the error message tells us to retry, after all :) ) I had hoped that the AWS SDK for .NET would handle this for us (indeed it handles retries for a lot of cases already - see https://github.com/aws/aws-sdk-net/blob/24a02b6c86d58753fbbd7158ab05ecadca5aec7b/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/DefaultRetryPolicy.cs for a comprehensive list) - but doesn't do 409, hence, it makes sense for us to handle that here. Also, thanks for adding Moq!

One thing I'd like to do a bit differently here is wrap each AWS operation in a retry loop of its own, rather than have one big retry loop around several operations as we have here. This would give us finer grained control over what's happening, and lets us use more specific error messages in the logs. Also I'd prefer to avoid recursion here. Ideally we could extract a decorator method that does the retries, and use this for each of creating the S3 bucket, setting the lifecycle config and creating the queue. 

If you have the time to make these updates I'd be more than happy to include this in the next release. Otherwise, I'm currently looking at releasing a v2.0.0 release of this plugin that supports NSB 6, and I'd be happy to include a retry policy for 409s as part of that work. Do you have plans to upgrade to NSB 6 any time soon?

ahofman:
I've also just noticed now that when you specify a name for an s3 bucket that is already in use by another account, you *also* get a 409 response. It doesn't make sense to retry this case, as it will always fail. We should check the "ErrorCode" field of the exception for "BucketAlreadyExists", and not retry in this case.

ahofman:
Have merged a fix as part of #59 , therefore will close this PR - thanks anyhow :)

chrissimon-au:
Hi @ahofman, sorry for the delay - no worries at all, glad you have a fix in now - it looks good!  One thing I added that you may like to include in a future update is the test for an existing matching lifecycle rule - as it's actually the adding of the lifecycle rule which is the primary culprit of the 409s, and especially in a cloud scaling scenario, the lifecycle rule will already be present the vast majority of times the service starts up, so removing the need to Put the lifecycle rule in the first place could improve startup times through removal of both the operation itself, and the 409 waits.

But either way - this should resolve the primary issue for us, so is appreciated!

Regarding NSB 6 - yep, we have been looking at an upgrade, but had been deferring as we'd like to keep using SQS :)

ahofman:
Good point @chrissimon-au , I have actually already implemented a check for an existing lifecycle rule in the NSB6 branch, as each endpoint was waiting several seconds due to the retries, as you mentioned. I'm happy to bring the same fix in to the NSB5 branch as well if you need it?

chrissimon-au:
No rush for the NSB5 branch - happy to wait until the NSB6 branch is ready to go.

Issue #87, opened 2017-07-01
Automate cleanup of SQS queues after acceptance tests run


ahofman:
Presently we need to manually delete (or purge) any SQS queue required by the Acceptance Tests in order to ensure they will pass. The Acceptance Tests should pass reliably and repeatedly with no manual steps.

ahofman:
Closed with #90

PR #90, opened 2017-07-01
Make the acceptance tests delete any created queues

ahofman:
Implement a single global cleanup method that deletes any queues created during the acceptance test run. 

Force a unique queue name prefix for each test run - to work around an SQS constraint where we cannot delete and then create a queue with the same name within 60 seconds.

Add some more code comments explaining how the queue name truncation works.

Part of #87


danielmarbach:
Close in favour of https://github.com/ahofman/NServiceBus.AmazonSQS/pull/92 ?

ahofman:
#92 didn't end up working as the mutator or pipeline behaviour can't stamp the test run id on outgoing subscribe and unsubscribe messages. 

Let's proceed with this approach for now - creating new queues in SQS is very quick, so creating new sets of queues for each test run may be OK.

If this turns out to be problematic, an alternative is to decorate the serializer, this is the approach used in ASQ: https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/216

PR #91, opened 2017-07-02
Acceptance tests cleanup based on test run ID

### LINKED ISSUE #87: Automate cleanup of SQS queues after acceptance tests run

SeanFeldman:
Fixes #87 
Alternative to #90


SeanFeldman:
@ahofman I don't know if it works or not as there's no visibility into build execution. Could you please verify it's working prior to merging? Thanks.

PR #92, opened 2017-07-02
Acceptance tests cleanup based on test run ID (#91)

ahofman:
Copy of #91 with a new branch in the ahofman repo, so I can test locally before merging to master. 

Alternative to #90


ahofman:
@SeanFeldman When running an acceptance test with a debugger attached, breakpoints in the new mutator aren't hit - The mutator needs to be registered somewhere I think?

SeanFeldman:
@ahofman pushed the missing part, sorry 🙂 
Neither sent a proper PR nor saw the movie ending. Multi-tasking is bad.

src/NServiceBus.AmazonSQS.AcceptanceTests/Infrastructure/TestIndependenceMutator.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace NServiceBus.AcceptanceTests.Infrastructure
+{
+    using System.Threading.Tasks;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.MessageMutator;
+
+    class TestIndependenceMutator : IMutateOutgoingTransportMessages
```
danielmarbach: @SeanFeldman can you remind me why we are doing this as a mutator again and not the outgoing physical part in a behavior?

ahofman:
@SeanFeldman @danielmarbach No luck I'm afraid - the mutator is never constructed nor invoked. The skip behaviour is set up and is invoked, but rejects all messages (because the mutator isn't active) and the tests fail. I tried this with MSMQ transport as well, same issue. Will investigate...

ahofman:
Not having much luck here - I've pushed commit d49fb03 to allow @SeanFeldman  or @danielmarbach 
 to test/debug locally - I've applied the test independence mutator and skip behavior to the Msmq transport, you should be able to place breakpoints in the mutator ctor and `MutateOutgoing` and observe them never being hit when debugging a test.

danielmarbach:
I've seen something similar in ASQ. We ended up writing the custom serializer. Let me investigate

SeanFeldman:
Update: PR #93 could address the issue.

ahofman:
Added 0293c56 as subscribe and unsubscribe messages were being filtered out - the stamping behaviour (and the mutator) weren't being applied to subscribe and unsubscribe messages. This fix makes the tests pass locally, but isn't perfect as it doesn't filter out any control messages from previous test runs. 

Perhaps we should reconsider #90 as a solution to the problem at hand?

ahofman:
Decided against - see #90

PR #93, opened 2017-07-03
Switching from mutator to behavior

danielmarbach:


danielmarbach:
@ahofman @SeanFeldman have you tried this?

SeanFeldman:
@ahofman we tried to run with MSMQ locally (@danielmarbach and myself) and it was working. Odd, but not sure what's going on. So this is an attempt to see if replacing a mutator with a behavior will help.

ahofman:
Thanks guys, will try this with fresh eyes tomorrow.

PR #105, opened 2017-07-18
Tweak atts

danielmarbach:


danielmarbach:
@mauroservienti I dropped the brackets commit and the timeout window for now since on TC the tests are not failing. We can always investigate more locally if the test continues to fail.

mauroservienti:
OK, cool. Still looks good 😃

PR #108, opened 2017-07-18
Cleanup

SimonCropp:


src/NServiceBus.AmazonSQS/Configure/SqsTransport.cs
```diff
@@ -10,20 +10,14 @@
     /// </summary>
     public class SqsTransport : TransportDefinition, IMessageDrivenSubscriptionTransport
     {
-        /// <summary>
-        /// Connection string error message.
-        /// </summary>
+        /// <inheritdoc />
```
SeanFeldman: @SimonCropp does `/// <inheritdoc />` work ootb or there's an additional stage/tool that is required to work with it? Asking because of this [discussion](https://visualstudio.uservoice.com/forums/121579-visual-studio-ide/suggestions/3745102-add-intellisense-support-for-the-inheritdoc-ta).

PR #112, opened 2017-07-19
Using Unicode.GetByteCount to determine size of payload

danielmarbach:
string.Length represent the number of chars in a string. [MSDN explanation](https://msdn.microsoft.com/en-us/library/system.string.length.aspx)

> Gets the number of characters in the current String object.

a string is a sequence of UTF-16 code units, namely chars

> Represents text as a sequence of UTF-16 code units.

Therefore the more accurate way of determining the size of the payload is to use either

`string.Length * sizeof(char)`

or use the framework helper method

``` 
System.Text.Encoding.Unicode.GetByteCount(string)
```

We would be constantly determining right now the message size to be half of its actual message size. So a normal message might be reported as 1244 bytes but it would be actually 2488 bytes.

### DIFF
--- a/src/NServiceBus.AmazonSQS/SqsMessageDispatcher.cs
+++ b/src/NServiceBus.AmazonSQS/SqsMessageDispatcher.cs
@@ -16,6 +16,7 @@
     using Logging;
     using Newtonsoft.Json;
     using Transport;
+    using static System.Text.Encoding;
 
     class SqsMessageDispatcher : IDispatchMessages
     {
@@ -35,7 +36,7 @@
                 {
                     var sqsTransportMessage = new SqsTransportMessage(unicastMessage.Message, unicastMessage.DeliveryConstraints);
                     var serializedMessage = JsonConvert.SerializeObject(sqsTransportMessage);
-                    if (serializedMessage.Length > 256 * 1024)
+                    if (Unicode.GetByteCount(serializedMessage) > MaximumPayloadSize)
                     {
                         if (string.IsNullOrEmpty(ConnectionConfiguration.S3BucketForLargeMessages))
                         {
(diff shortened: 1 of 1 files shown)


SimonCropp:
@danielmarbach LGTM. but is there a way to force a failing test prior to this change?

danielmarbach:
Yeah good idea, sorry I should have thought about that. Will add

ahofman:
Nice find :) Could use [this](https://github.com/Particular/NServiceBus.AmazonSQS/blob/develop/src/NServiceBus.AmazonSQS.AcceptanceTests/When_sending_messages_larger_than_256kb.cs) as a starting point for the test? (just tweak PayloadSize)

danielmarbach:
Apparently, SQS on the server side does something like a string length check only. I was playing around with wolfram alpha to come up with a reasonable size for the message. The problem is that base64 adds 4/3 on the payload, so to have a realistic message size you have to use the following formula

```
(x*1024/3)*4=DesiredSize*1024
```

So for example to have a payload (without taking headers into consideration) that is larger than 256 KB you can solve the following equation

image

https://www.wolframalpha.com/input/?i=(x*1024%2F3)*4%3D257*1024

which leads to 

```
x = 192.75
```

So the byte array size can be around 193 bytes which then already brings you over the payload limit. I was assuming that the actual payload size on the server would take into account the char size and that's why it sent in this PR. If it would be the case then a payload size of 129 KB should already fail on the client but pass through our test. 

https://www.wolframalpha.com/input/?i=(x*1024%2F3)*4%3D129*1024

which leads to 

```
x = 96.75
```

but choosing a payload of 97 KB did not break the call to the SDK API so I'm withdrawing this PR. 

@Particular/aws-maintainers should we change the existing acceptance tests to better reflect that the payload can be smaller than 256 KB and it will still fall back to using S3 because of the overhead of our base64 encoding?

PR #115, opened 2017-07-19
Async tweaks

SimonCropp:


danielmarbach:
@SimonCropp I took the liberty and rebased this since I merged the other changes on the dispatcher before this one. I will merge when green since it is a good step towards faster dispatch. I would even go so far and remove the try catch since having an await just for the log doesn't really add much because the pipeline will log any errors anyway.

PR #130, opened 2017-07-26
Ignore machine name

### LINKED ISSUE #128: Duplicate events published when scaling out subscribers
Affects 1.x releases (for NSB v5).

The ReplyToAddress is given in the form "Endpoint@MachineName". When subscribers are scaled out across multiple machines, Publishers end up with one subscription per machine per subscriber. The publisher publishes one event per subscribing machine, when it should only publish one event per subscribing endpoint. 

Consider using `Address.IgnoreMachineName()` as the Sql transport does.

ahofman:
Ignore machine name when addressing. 
Fixes #128 

Users affected by this issue will need to manually fix any existing subscription stores that have multiple entries for the same endpoint (but with different machine names) - hence this is not necessarily a breaking change, rather some manual patching is required to take advantage of the fix.

### DIFF
--- a/src/NServiceBus.AmazonSQS/Configure/SqsTransport.cs
+++ b/src/NServiceBus.AmazonSQS/Configure/SqsTransport.cs
@@ -15,6 +15,8 @@
 
 		protected override void Configure(BusConfiguration config)
 		{
+            Address.IgnoreMachineName();
+
 			config.EnableFeature<SqsTransportFeature>();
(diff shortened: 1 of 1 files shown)


mauroservienti:
is the file spacing weird due to a mix of tabs and spaces?

ahofman:
Yeah the whitespace in the 1.3 branch is a mess :(

mauroservienti:
no problems 😃 The content looks good 😉

PR #136, opened 2017-08-08
SDK bump

danielmarbach:
https://github.com/aws/aws-sdk-net/issues/720 is clarified

Changed according to

> Use the [3.3, 3.4) range. Although I would suggest setting the min in the range to the latest version at the time of your release. There have been a lot of fixes done to the SDK since 3.3 came out and by default NuGet will pull in the lowest compatible dependency that satisfies all the dependencies.

### DIFF
--- a/packaging/nuget/nservicebus.amazonsqs.nuspec
+++ b/packaging/nuget/nservicebus.amazonsqs.nuspec
@@ -15,10 +15,10 @@
     <copyright>$copyright$</copyright>
     <tags>$tags$</tags>
     <dependencies>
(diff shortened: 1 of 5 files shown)


danielmarbach:
@ahofman @mauroservienti changed according to guidelines of SDK team

> Use the [3.3, 3.4) range. Although I would suggest setting the min in the range to the latest version at the time of your release. There have been a lot of fixes done to the SDK since 3.3 came out and by default NuGet will pull in the lowest compatible dependency that satisfies all the dependencies.

PR #138, opened 2017-08-09
When shutting down messages fetched should not be marked as poison

danielmarbach:
When shutting down messages fetched should not be marked as poison (happens with S3)
Early opt out when cancellation is requested.

### DIFF
--- a/src/NServiceBus.AmazonSQS/MessagePump.cs
+++ b/src/NServiceBus.AmazonSQS/MessagePump.cs
@@ -108,6 +108,11 @@
 
                     var tasks = receiveResult.Messages.Select(async message =>
                     {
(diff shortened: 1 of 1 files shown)


SimonCropp:
@danielmarbach assume it is too much hassle to have a test that verifies this? if so i am ok with a merge

SimonCropp:
also should this be in the milestone so it shows in the release notes?

danielmarbach:
@SimonCropp I could not come up with a good way to test this scenario because it would require augmenting the pump to halt in the specific case. Maybe a unit test could reproduce it somehow. Do you want me to take a stab?

SimonCropp:
Nope ;)

PR #141, opened 2017-08-29
run acc tests on newest nsb

SimonCropp:


SimonCropp:
i have a suspicion whit build will fail

mauroservienti:
it's green 😄 merge?

PR #146, opened 2017-09-14
remove redundant code

SimonCropp:


src/Tests/APIApprovals.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System.IO;
+﻿#if (net452)
```
danielmarbach: @SimonCropp why the braces around it? Never seen that syntax. More used to https://github.com/Particular/NServiceBus/blob/72075be388c49d6d229f84b939db4f99f6007c2b/src/NServiceBus.Core.Tests/API/APIApprovals.cs#L1

danielmarbach:
Good changes. I'm OK that this was merged without a second pair of eyes since these changes are purely related to minor cleanups and well isolated

PR #154, opened 2017-10-09
make ttl a timespan

### LINKED ISSUE #149: MaxTTLDays(int) API should be MaxTTL(TimeSpan)
The TTL can currently only be overriden as days while AmazonSQS supports this to be configured in seconds with a minimum of 60 seconds.

SimonCropp:
fixes #149


src/Tests/APIApprovals.ApproveSqsTransport.approved.txt
```diff
@@ -21,6 +21,9 @@ namespace NServiceBus
     public class static SqsTransportSettings
     {
         public static NServiceBus.TransportExtensions<NServiceBus.SqsTransport> CredentialSource(this NServiceBus.TransportExtensions<NServiceBus.SqsTransport> transportExtensions, NServiceBus.SqsCredentialSource credentialSource) { }
+        public static NServiceBus.TransportExtensions<NServiceBus.SqsTransport> MaxTtl(this NServiceBus.TransportExtensions<NServiceBus.SqsTransport> transportExtensions, System.TimeSpan maxTtl) { }
```
danielmarbach: Why `MaxTtl` and not `MaxTTL`?

danielmarbach:
@SimonCropp It seems two transport tests are failing. But as it looks the transport tests seem to be a bit flaky already on develop. Not sure how to proceed here and if we should take a deeper look into the transport test failures. 

I would actually prefer to name it `MaxTTL` but other than that I'm good

mauroservienti:
On the fence about the variable naming, other then that it's good to go.

danielmarbach:
Given that [ASB uses TTL](https://docs.particular.net/nservicebus/messaging/databus/azure-blob-storage?version=absdatabus_2) and [rabbit diagram uses TTL](https://docs.particular.net/transports/rabbitmq/delayed-delivery?version=rabbit_5#how-it-works-delay-levels) as well and we abbreviate time to be received with [TTBR ](https://docs.particular.net/nservicebus/messaging/discard-old-messages?version=core_7)I think we should rename it to TTL.

src/NServiceBus.AmazonSQS/Configure/SqsTransportSettings.cs
```diff
@@ -31,23 +31,26 @@ public static TransportExtensions<SqsTransport> Region(this TransportExtensions<
...
+        /// </remarks>
         /// <param name="transportExtensions"></param>
-        /// <param name="maxTtlDays">The max TTL in days. Must be a value between 1 and 14.</param>
-        public static TransportExtensions<SqsTransport> MaxTTLDays(this TransportExtensions<SqsTransport> transportExtensions, int maxTtlDays)
+        /// <param name="maxTtl">The max TTL in days. Must be a value between 0 and not greater than 14 days.</param>
+        public static TransportExtensions<SqsTransport> MaxTtl(this TransportExtensions<SqsTransport> transportExtensions, TimeSpan maxTtl)
         {
-            if (maxTtlDays <= 0 || maxTtlDays > 14)
+            var maxDays = TimeSpan.FromDays(14);
+            if (maxTtl <= TimeSpan.Zero || maxTtl > maxDays)
```
danielmarbach: > The default message retention period is 4 days. However, you can set the message retention period to a value from 60 seconds to 1,209,600 seconds (14 days) 

http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-lifecycle.html

danielmarbach:
@mauroservienti @SimonCropp I added my proposal. 

@SimonCropp brought up a point in slack

> all upper case acronyms make variable and param naming difficult. eg should the param name be tTL?

I'm ok to follow @SimonCropp lead if we decide to deviate from the samples I linked in https://github.com/Particular/NServiceBus.AmazonSQS/pull/154#issuecomment-335086488 or we could just spell it out

PR #155, opened 2017-10-09
Introduce SQS client factory and S3 factory

### LINKED ISSUE #129: Setting the credentials in production
Whats the best way to set up the credentials in non-aws production environment? Setting the access and secret key environmental variable is not ideal.

danielmarbach:
Closes #129

Introduces a possibility to specify a client factory for the SQS and the S3 client creation. 

## Documentation

https://github.com/Particular/docs.particular.net/pull/3337

## SQS Client Factory

Example

```
transport.ClientFactory(() => new AmazonSQSClient(
   new AmazonSQSConfig {
      RegionEndpoint = "ap-southeast-2"
   });
```

## S3 Client Factory

Example

```
transport.ClientFactory(() => new AmazonS3Client(
   new AmazonS3Config {
      RegionEndpoint = "ap-southeast-2"
   });
```


mauroservienti:
Given all the magic the the AWS client put in place to support many configuration sources, why enforcing people on providing a factory that probably in the majority of cases would be something like:

```
transportSettings.ClientFactory(() => new AmazonSQSClient());
```

I'd rather make the above the default behavior, behavior the if needed can be easily overwritten by users.

danielmarbach:
@mauroservienti that would be good for me too let's wait for @SimonCropp to chime in.

SimonCropp:
> I'd rather make the above the default behavior, behavior the if needed can be easily overwritten by users.

@mauroservienti yep that sounds reasonable

src/AcceptanceTests/SetupFixture.cs
```diff
@@ -35,17 +35,19 @@ public void OneTimeSetUp()
...
-                catch (Exception ex)
-                {
-                    Console.WriteLine($"Exception when deleting queue: {ex}");
+                    try
+                    {
+                        await sqsClient.DeleteQueueAsync(queueUrl).ConfigureAwait(false);
+                    }
+                    catch (Exception ex)
+                    {
+                        Console.WriteLine($"Exception when deleting queue: {ex}");
```
SimonCropp: why Console.WriteLine  ?

danielmarbach: This was already there, I think it is done because then it is visible on TC in the build logs

danielmarbach:
Added the [doc PR WIP](https://github.com/Particular/docs.particular.net/pull/3337) to the description

danielmarbach:
@mauroservienti @SimonCropp as soon as the PR is merged I will continue with the doc PR.

src/NServiceBus.AmazonSQS/Configure/S3Settings.cs
```diff
@@ -0,0 +1,93 @@
...
+    using System;
+    using System.Linq;
+    using Amazon.S3;
+    using Configuration.AdvancedExtensibility;
+    using Settings;
+
+    /// <summary>
+    /// 
+    /// </summary>
+    public class S3Settings : ExposeSettings
```
SimonCropp: needs some xml comments.

PR #158, opened 2017-10-11
Move tests into NServiceBus.AcceptanceTests

danielmarbach:


danielmarbach:
It seems the transport tests are sometimes timing out. Need to investigate

SimonCropp:
@danielmarbach but that should not block this PR?

PR #160, opened 2017-10-12
Remove fluent API from S3 config and make settings mandatory

danielmarbach:


mauroservienti:
Scratch the above, @danielmarbach explained me the details. Approving.

PR #163, opened 2017-10-30
Project cleanup

bording:
This cleans up some things related to moving to the new project system and Particular.Packaging. I also noticed a few other misc things that seemed to be missing.

One other thing I changed here was that I didn't see a reason to be referencing AWSSDK.Core directly since both the S3 and SQS packages reference that as well.

### DIFF
--- a/src/AcceptanceTests/AcceptanceTests.csproj
+++ b/src/AcceptanceTests/AcceptanceTests.csproj
@@ -1,20 +1,23 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
   <PropertyGroup>
     <TargetFrameworks>net452;netcoreapp2.0</TargetFrameworks>
     <SignAssembly>true</SignAssembly>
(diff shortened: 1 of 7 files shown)


danielmarbach:
Thanks @bording for the PR!. Hugely appreciated. We will look into this and why the build is failing as soon as possible.

mauroservienti:
2 tests are failing due to test execution timeout, the 3rd one fails because there is no S3 bucked configured. It's strange that at least the last one works on Windows though. I'll investigate further

mauroservienti:
is it also OK that we're running tests against an expired trial license?

> No valid license could be found. Falling back to trial license with start date '2017-08-07'.
> Trial for the Particular Service Platform has expired.

It's for sure not the cause of the failure

bording:
@Particular/aws-maintainers This PR still needs to be merged. Have you had time to investigate the test failures? I don't think they're related to the changes here.

mauroservienti:
I’m still trying to find the time to have a look, the failure seems to be totally unrelated to che changes.

SzymonPobiega:
@mauroservienti I never seen something like this :-(

kbaley:
@mauroservienti This might me due to permissions but probably not likely if the tests fail intermittently. I'll ping you Monday morning my time to investigate.

mauroservienti:
Thanks @kbaley

PR #166, opened 2017-11-14
Updates env var names

mauroservienti:
As discovered environment variables containing a `.` in the name are not supported on Linux, I already created new variables on TC, tested them both on TC and locally. Changed in the code via https://github.com/Particular/NServiceBus.AmazonSQS/pull/163/commits/6c69f01e51f060fd132c8370e843f2d31f3f2034. The last thing is to update the readme.

Please review and merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,9 +15,9 @@ To run the tests, the Access Key ID and Secret Access Key of an AWS IAM account 
 
 The transport can be configured using the following environment variables:
 
- * **NServiceBus.AmazonSQS.Region** corresponds to the [Region](https://docs.particular.net/transports/sqs/configuration-options#region) parameter. Default is no region set.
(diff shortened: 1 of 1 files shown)


MarcinHoppe:
Is this consistent across all released versions?

danielmarbach:
@MarcinHoppe previous versions will use the `.` containing variables, so no it is not consistent. As long as TC supports both we have to remember to have both versions available on our machines when we run the ATTs locally for a support branch

MarcinHoppe:
Maybe we could list env variables per version to keep it precise?

danielmarbach:
@MarcinHoppe the readme reflects the state of the repo of the current branch, so if you go to a support branch the readme will reflect the state of the env variable at that time. Isn't that enough?

MarcinHoppe:
@danielmarbach Thanks, my mind went dark for some time. Scratch my last comment :).

PR #172, opened 2017-11-16
require windows

yvesgoeleven:
Startup check that requires the windows os

### DIFF
--- a/src/NServiceBus.AmazonSQS/Configure/RequiresWindows.cs
+++ b/src/NServiceBus.AmazonSQS/Configure/RequiresWindows.cs
@@ -1,6 +1,6 @@
 ﻿namespace NServiceBus
 {
-    using System.Runtime.InteropServices;
+    using System;
     using System.Threading.Tasks;
     using Transport;
 
@@ -8,7 +8,7 @@
     {
         public Task<StartupCheckResult> Validate()
         {
-            var isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+            var isWindows = Environment.OSVersion.Platform == PlatformID.Win32NT;
             return Task.FromResult(isWindows ? StartupCheckResult.Success : StartupCheckResult.Failed("Due to a bug in the AWS SDK on linux, Windows is the only supported platform."));
         }
     }
--- a/src/NServiceBus.AmazonSQS/Configure/RequiresWindows.cs
+++ b/src/NServiceBus.AmazonSQS/Configure/RequiresWindows.cs
@@ -1,0 +1,15 @@
+﻿namespace NServiceBus
+{
+    using System.Runtime.InteropServices;
+    using System.Threading.Tasks;
+    using Transport;
+
+    class RequiresWindows
+    {
+        public Task<StartupCheckResult> Validate()
+        {
+            var isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+            return Task.FromResult(isWindows ? StartupCheckResult.Success : StartupCheckResult.Failed("Due to a bug in the AWS SDK on linux, Windows is the only supported platform."));
+        }
+    }
+}
--- a/src/NServiceBus.AmazonSQS/Configure/TransportInfrastructure.cs
+++ b/src/NServiceBus.AmazonSQS/Configure/TransportInfrastructure.cs
@@ -61,7 +61,7 @@
             return new TransportReceiveInfrastructure(
                 CreateMessagePump,
                 CreateQueueCreator,
-                () => Task.FromResult(StartupCheckResult.Success));
+                () => new RequiresWindows().Validate());
         }
 
         public override TransportSendInfrastructure ConfigureSendInfrastructure()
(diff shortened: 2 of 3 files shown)


yvesgoeleven:
Linux build failed as expected, paused it for now

mauroservienti:
Should we disable Linux tests in the meanwhile

yvesgoeleven:
@mauroservienti  I did

mauroservienti:
:+1:

src/NServiceBus.AmazonSQS/Configure/TransportInfrastructure.cs
```diff
@@ -61,7 +61,7 @@ public override TransportReceiveInfrastructure ConfigureReceiveInfrastructure()
             return new TransportReceiveInfrastructure(
                 CreateMessagePump,
                 CreateQueueCreator,
-                () => Task.FromResult(StartupCheckResult.Success));
+                () => new RequiresWindows().Validate());
```
bording: Do we want to allow send-only endpoints on linux? If not, this startup check should be on `ConfigureSendInfrastructure` instead.

src/NServiceBus.AmazonSQS/Configure/RequiresWindows.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    using System.Runtime.InteropServices;
+    using System.Threading.Tasks;
+    using Transport;
+
+    class RequiresWindows
+    {
+        public Task<StartupCheckResult> Validate()
+        {
+            var isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
```
bording: While using `RuntimeInformation.IsOSPlatform` is the better, preferred API for several reasons, including that it can distinguish between linux and macOS, it might be simpler to use `Environment.OSVersion.Platform == PlatformID.Win32NT` instead. This would avoid the need for the `System.Runtime.InteropServices.RuntimeInformation` package dependency.

src/NServiceBus.AmazonSQS/Configure/TransportInfrastructure.cs
yvesgoeleven: As the bug is receive only, I'd say 'why not'?

src/NServiceBus.AmazonSQS/Configure/RequiresWindows.cs
yvesgoeleven: Is there a significant downside to having the dependency on `System.Runtime.InteropServices.RuntimeInformation`?

src/NServiceBus.AmazonSQS/Configure/TransportInfrastructure.cs
bording: It might be an easier story to say that we just don't support it on linux at all right now vs the more complicated story of "Not supported on linux, unless it's a send-only endpoint."

src/NServiceBus.AmazonSQS/Configure/RequiresWindows.cs
bording: Given that the need for `RuntimeInformation.IsOSPlatform` would be temporary, we'd be adding a dependency that is also temporary. You could argue that removing a dependency would be a breaking change, especially given how transitive NuGet dependencies work these days, so we'd either have to keep the dependency around even though we don't need it, or release a new major just to let linux work once the bug is fixed.

src/NServiceBus.AmazonSQS/Configure/TransportInfrastructure.cs
danielmarbach: I agree but on the other hand this would at least allow for example to host ASP.NET Core with Sendonly endpoints on Linux and the backend system on NET Core with windows until resolved.

src/NServiceBus.AmazonSQS/Configure/RequiresWindows.cs
danielmarbach: Good argument @bording and nice insights. I suggest we switch to @bording suggestion?

yvesgoeleven:
Removed the dependency on System.Runtime.InteropServices.RuntimeInformation, linux build still failing as expected, will disable it again.

src/NServiceBus.AmazonSQS/Configure/TransportInfrastructure.cs
yvesgoeleven: I suggest the @Particular/aws-maintainers take a decission here on what to allow

andreasohlund: Assuming we're going to disable the linux tests completely wouldn't this mean that we have a "supported" feature (send only on linux) where we have zero acceptance tests covering it? 

Can we do something in the interim to cover the gap? manual smoke test with a sample etc?

danielmarbach: @andreasohlund yeah that is a good point but given the dispatcher logic is super simple I would not be too worried about it since we test send and receive on the full framework.

PR #173, opened 2017-12-07
Disposing S3Client and SqsClient when transport stops

danielmarbach:


MarcinHoppe:
@danielmarbach Could we use Janitor to do it for us?

danielmarbach:
We could yeah but I'm leaning more towards not using that magic especially because the transport infrastructure controls the clients and not the pump and the URL cache that uses it. Not sure how Janitor would handle this scenario.

PR #185, opened 2017-12-18
Prepare for final beta release

bording:
@Particular/aws-maintainers There are still errors related to different versions of AWS S3 analyzer assemblies because the test projects are still using wildcards and are pulling in a newer version of the package.

There are a couple ways to solve this. If using a newer client in the tests is desired, the package references could be set up to exclude the analyzers for the test projects.

### DIFF
--- a/src/AcceptanceTests/AcceptanceTests.csproj
+++ b/src/AcceptanceTests/AcceptanceTests.csproj
@@ -14,7 +14,7 @@
   <ItemGroup>
     <PackageReference Include="AWSSDK.S3" Version="3.*" />
     <PackageReference Include="AWSSDK.SQS" Version="3.*" />
-    <PackageReference Include="NServiceBus.AcceptanceTests.Sources" Version="7.0.0-*" />
+    <PackageReference Include="NServiceBus.AcceptanceTests.Sources" Version="7.0.0-beta0012" />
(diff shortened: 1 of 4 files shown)


bording:
Looking at options for solving the analyzer problem I mentioned above, I saw that there is an issue open to fix the problem in the AWS packages: https://github.com/aws/aws-sdk-net/issues/751

The documentation for package references indicate that it should be possible to exclude analyzers from a package by adding `ExcludeAssets="analyzers"` to the package reference, but it looks like this isn't actually working yet: https://github.com/dotnet/sdk/issues/1212

So, the only way to make them go away right now would be to reference the same version of the AWS packages from the main project and the test project. 

However, the errors aren't actually causing the build to break, so just leaving them there for now seems like an option as well.

danielmarbach:
Let's reference the same version for now. Can you add it or shall I do it?

bording:
The API approval test is failing because of some problem with Mono.Cecil. Looks like we'll need to investigate what's causing that.

bording:
Looks like the 2.3.1 release of Fody was the culprit. Going to an older release or the brand new 2.3.2 release has fixed the problem.

Fody must have been corrupting the PDB in some way, causing the API approval test to fail when it tried to read the PDB.

danielmarbach:
Thanks @bording

PR #197, opened 2018-02-27
Native Delayed Deliveries support

mauroservienti:
@danielmarbach and myself kicked off Native Delayed Deliveries support implementation today after with @bording and @SeanFeldman we decided how to design the new feature.

We started by designing some acceptance tests to validate behavior of send-only endpoints in various scenarios:

1. Send-Only: Sending delayed messages when sender and receiver are properly configured:
   * Should deliver messages if delay is below AWS threshold (default 15 minutes);
   * Should deliver messages if delay is above AWS threshold (default 15 minutes);
1. Send-Only: Sending delayed messages when sender is not properly configured:
   * Should deliver messages only if delay is below AWS threshold, by using the legacy deferred delivery mode
   * Should fail to deliver messages if delay above AWS threshold
1. Send-Only: Sending delayed messages when receiver is not properly configured:
   * Should deliver messages only if delay is below AWS threshold, by using the legacy deferred delivery mode
   * Should fail to deliver messages if delay above AWS threshold

The question @danielmarbach and myself raised is: do we need to test also non-send-only endpoints as senders? Or are we good enough?


src/AcceptanceTests/DelayedDelivery/SendOnly_Sending_when_sender_and_receiver_are_properly_configured.cs
```diff
@@ -54,14 +54,11 @@ public class SendOnly_Sending_delayed_messages_when_sender_and_receiver_are_prop
...
 
-                    c.SentAt = DateTime.UtcNow;
+                    c.MessageSent = true;
                 }))
                 .WithEndpoint<Receiver>()
-                .Done(c => c.Received)
+                .Done(c => c.MessageSent)
                 .Run();
-
-            Assert.GreaterOrEqual(context.ReceivedAt - context.SentAt, delay, "The message has been received earlier than expected, we're so good!");
```
danielmarbach: Why was this assertion remove?

src/AcceptanceTests/DelayedDelivery/SendOnly_Sending_when_sender_and_receiver_are_properly_configured.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            Assert.GreaterOrEqual(context.ReceivedAt - context.SentAt, delay, "The message has been received earlier than expected, we're so good!");
+            Assert.AreEqual(payload, context.Payload, "The received payload doesn't match the sent one. BAD BAD BAD");
+        }
+
+        [Test]
+        public async Task Should_send_message_if_above_threshold()
+        {
+            var payload = "some payload";
+            var delay = TimeSpan.FromMinutes(16);
```
danielmarbach: Why minutes? The idea was that we make the threshold configurable and then we can do any arbitrary seconds delay over fifo

SeanFeldman:
@danielmarbach @mauroservienti 
@bording and I have continued on the implimentation and have done the following:
- Updated acceptance tests with a few modification. Specifically, changing sending tests with delays more than 15 minutes to verify send operation is successful without delivery verification.
- Obsoleted with warning public API `.NativeDeferral()` that is doing nothing.
- Introduced a new public API to enable delayed delivery (`.UnrestrictedDurationDelayedDelivery()`.
- Modified dispatcher to handled delayed delivery over 15 minutes using FIFO queues.
- Modified `QueueCreator` to handle delayed delivery queues creation.
    - Ran into an issue we should address is a separate PR for SQS v3 - `transportInfrastructure.ToTransportAddress()` doesn't use `QueueNameHelper`. Instead, `QueueNameHelper` is used everywhere.

At this point we ran into a tinny tiny issue: on a FIFO queue, a delay cannot be set on a message level. Instead it needs to be declared on the queue level. This means we cannot handle delays less than 15 minutes on FIFO queues. We need to discuss this roadblock and figure out how to move forward during our scheduled sync for tomorrow.

src/AcceptanceTests/DelayedDelivery/SendOnly_Sending_when_sender_and_receiver_are_properly_configured.cs
SeanFeldman: We want to avoid configuring transport to accomodate test values.
In addition to that, we'd not be testing anything above 15 minutes anyways using ATTs, so better to verify that send is a valid send. Send and delivery with over 15 minutes would need to be a manual smoke test.

mauroservienti: Well, we could configure the `*-delays.fifo` queue to have a smaller delay than 15 minutes, say 15 seconds and use that as a test baseline

SeanFeldman:
> I still think those scenarios would have been good to have as tests

What is it that those tests have covered that we didn't have with the already existing ones?

danielmarbach:
I personally would like to bring the tests in again that were removed in https://github.com/Particular/NServiceBus.AmazonSQS/pull/197/commits/f24c0b81e00f8d6a33ccb0d1daf57f5eeb2d4afb Other than that this is good to go.

danielmarbach:
@mauroservienti @bording @SeanFeldman I aligned the delayed delivery queue TTL with the main input queue TTL https://github.com/Particular/NServiceBus.AmazonSQS/pull/197/commits/cbec845854de8bf96031b564739a6603e8f0adae. I see no reason why it should be different

danielmarbach:
Today we decided to revert https://github.com/Particular/NServiceBus.AmazonSQS/pull/197/commits/22520ba6a2d28d408c30706faf1db98174d50a2c by rewriting history when we target the release branch. @bording will do it

PR #204, opened 2018-03-20
Remove linux restriction

SeanFeldman:
Remove Linux restriction.


mauroservienti:
Marked as wip since we probably want to fix tests and build configuration in the same PR

SeanFeldman:
There were two timing out tests on Linux. Re-running tests fixed that. We'll need to increase wait time.
Where do I find ATT max execution time?

SeanFeldman:
Answered my own question 😆 
`ConfigureEndpointSqsTransport : IConfigureEndpointTestExecution`

SeanFeldman:
@mauroservienti looks like we'll swtich 🙂 
Could you please review and see if I've missed anything.

SeanFeldman:
> Marked as wip since we probably want to fix tests and build configuration in the same PR

Build configuration is enabled and tests max execution time increased.

src/AcceptanceTests/ConfigureEndpointSqsTransport.cs
```diff
@@ -27,7 +27,7 @@ public Task Configure(string endpointName, EndpointConfiguration configuration,
                 }
             }
 
-            settings.TestExecutionTimeout = TimeSpan.FromSeconds(20);
+            settings.TestExecutionTimeout = TimeSpan.FromSeconds(20 * 2);
```
danielmarbach: Why doing math instead of 40 seconds?

How about setting it to 40 seconds and then write a comment that the value has to be greater than 30 secs (visibility timeout)?

mauroservienti: I have no preference, up to @SeanFeldman to decide 😄

PR #205, opened 2018-03-23
Update Newtonsoft.Json 11.0.1

danielmarbach:
Based against release-3.2.0 in preparation for a new minor release


danielmarbach:
.NET Core builds are supposed to fail

Issue #213, opened 2018-04-04
Wire-compatibility issue between NSB5 (SQS Transport 1.3.1) and NSB 6 (SQS Transport 2.x/3.x)


chrissimon-au:
In v1.3.1, the SQS Transport had a transport message property `TimeToBeReceived`.  It seems that in SQS Transport 2.x / 3.x, this has been moved to a standard NSB header.

This breaks wire-compatibility:

1.3.1 has logic that assumes that a message is expired and is to be discarded if the `TimeToBeReceived` header is not equal to `TimeSpan.MaxValue` - which is won't be if it's not present.  See https://github.com/Particular/NServiceBus.AmazonSQS/blob/hotfix-1.3.1/src/NServiceBus.AmazonSQS/SqsDequeueStrategy.cs#L168.

(Even worse, there is a bug in the discard approach: it does not delete the message from the queue, so the message becomes a poison message that is infinitely re-received and released back to the queue.)

This impacts on an upgrade pathway - we had been assuming that we could upgrade endpoints to NSB6 one-by-one in any sequence and rely on wire-protocol compatibility to ensure interoperation between NSB5 and NSB6 endpoints.  This issue means that messages can flow from NSB5 endpoints -> NSB6 endpoints, but not from NSB6 endpoints -> NSB5. (And even then, only if `TimeToBeReceived` is not a required feature).

While we could adjust our upgrade plans to accommodate the necessary sequencing of upgrade, it would be great if this could be addressed with an update to the transport - either by having v3.x add `TimeToBeReceived` onto the transport message (perhaps as an optional config option) or by releasing a 1.3.x update to make it forward compatible with the new wire protocol.

Thanks!

mauroservienti:
@Particular/aws-maintainers does this qualify as a `patch`?

mauroservienti:
@chrissimon-au thanks for reporting this, and sorry for the the pain we're causing. We'll investigate options and get back to you soon.

chrissimon-au:
No problem - for what it's worth, I agree with the design decision to move the `TimeToBeReceived` property to a header - it's not an SQS-specific property, and it makes sense that the `SqsTransportMessage` would only have properties specific to transporting over SQS.

It's just a shame about the backwards compatibility implications!

chrissimon-au:
[...]

PR #214, opened 2018-04-26

boblangley:
Connects to #213

The intent is to:


boblangley:
@SeanFeldman this is truly WIP, but represents the overall approach I am suggesting to fix. Can you take a look and we can talk to see if this looks like a good way to fix the issue? @mauroservienti if you have time it would be great to get your input as well.

src/NServiceBus.AmazonSQS/MessagePump.cs
```diff
@@ -258,15 +258,17 @@ void ProcessMessages(List<Message> receivedMessages, List<Task> concurrentReceiv
...
-        static bool IsMessageExpired(Message receivedMessage, IncomingMessage incomingMessage)
+        static bool IsMessageExpired(Message receivedMessage, TransportMessage incomingMessage)
         {
-            if (!incomingMessage.Headers.TryGetValue(TransportHeaders.TimeToBeReceived, out var rawTtbr))
+            incomingMessage.Headers.TryGetValue(Headers.MessageId, out var messageId);
+
+            if (!TimeSpan.TryParse(incomingMessage.TimeToBeReceived, out var timeToBeReceived))
             {
-                return false;
+                throw new Exception($"Unable to parse TTBR value {incomingMessage.TimeToBeReceived} for message {messageId}.");
```
SeanFeldman: **`TTBR`** abbreviation should be spelled out

boblangley:
@SeanFeldman You are faster than I expected, I realized I don't need to touch `MessagePump` at all so I reverted all of those changes.

mauroservienti:
Thinking out loud: this is taking the approach of making 2.x and 3.x backward compatible, fine. What if we simply make 1.x forward compatible by making sure that the 2 properties value are stored in the headers?

danielmarbach:
Additionally doesn't this take more serialization space now because the
headers are populated to the TM which is serialized and deserialized
including those properties?

Originally what we discussed in slack was to patch 1.1.x to add to what
@mauroservienti said

Mauro Servienti <notifications@github.com> schrieb am Do., 26. Apr. 2018,
06:50:

boblangley:
@mauroservienti / @danielmarbach that is another way of approaching this.

To ensure success with an upgrade we will require them to update **all** of their endpoints that talk to the new 2+x endpoint, and then the new endpoints will be compatible. This is a _forwards_ compatibility argument and is a break from our usual policy in this regard which is _backwards_ compatibility.

I don't care much for the forward compatibility approach.

SeanFeldman:
> Additionally doesn't this take more serialization space now because the headers are populated to the TM which is serialized and deserialized including those properties?

While this is true, we're not talking about massive amount of data - TimeSpan and 80 charchters for queue name. This probably outweights the consistency with backwards compatibility of the platform.

src/NServiceBus.AmazonSQS/TransportMessage.cs
```diff
@@ -29,15 +29,51 @@ public TransportMessage(OutgoingMessage outgoingMessage, List<DeliveryConstraint
...
 
         public string S3BodyKey { get; set; }
+
+        public string TimeToBeReceived
+        {
+            get
+            {
+                if (timeToBeReceived == null)
+                {
+                    timeToBeReceived = Headers.ContainsKey(TransportHeaders.TimeToBeReceived) ? Headers[TransportHeaders.TimeToBeReceived] : TimeSpan.MaxValue.ToString();
```
SeanFeldman: A backing field was used for `TimeToBeReceived`, but not for `ReplyToAddress`. Is it necessary?

danielmarbach:
Regarding https://github.com/Particular/NServiceBus.AmazonSQS/pull/214#issuecomment-384849820

Is there no other way? Could we add a json ignore? We are talking about a transport with limited msg size and I think we should not make such decisions without first outlining other possibilities

boblangley:
Another possibility, one I started with but never pushed, was to have a transport setting determine if the property is serialized or not to indicate you need backwards compatibility with 1.x transports.

boblangley:
Here are some commits with a configurable backwards compatible option. Right now it is not backwards compatible by default.

mauroservienti:
Haven’t reviewed the changes yet. Since we’re making v3 backward compatible, people actually running on v1 can deploy newer endpoints with the compatibility settings etc to true, once the migration process is completed and all endpoints moved away from v1 in a subsequent deployment the compatibility setting can be set back to its default, false. This also means that in an hypothetical vNext we can drop v1 compatibility code. The cons of this approach is that these changes need to be applied to v2 as well.

Correct?

boblangley:
@mauroservienti Correct.

src/NServiceBus.AmazonSQS/Configure/SqsTransportSettings.cs
```diff
@@ -170,6 +170,19 @@ public static TransportExtensions<SqsTransport> NativeDeferral(this TransportExt
             return transportExtensions;
         }
 
+        /// <summary>
+        /// Configures the SQS transport to be compatible with 1.x versions of the transport.
+        /// </summary>
+        /// <param name="transportExtensions"></param>
+        /// <param name="use">Set to true to use allow the endpoint to send messages to endpoints using version 1.x of the transport; false otherwise.</param>
+        /// <returns></returns>
+        public static TransportExtensions<SqsTransport> V1CompatibilityMode(this TransportExtensions<SqsTransport> transportExtensions, bool use = true)
```
mauroservienti: If by default the endpoint is not backward compatible what’s the point of setting it to `false`?
Shouldn’t this extension method simply be something like `EnableV1CompatibilityMode()` with no arguments? (Or `RunInV1CompatibilityMode()`)

src/NServiceBus.AmazonSQS/ReducedPayloadContractResolver.cs
```diff
@@ -0,0 +1,24 @@
+﻿namespace NServiceBus.AmazonSQS
+{
+    using System.Collections.Generic;
+    using System.Reflection;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Serialization;
+    class ReducedPayloadContractResolver : DefaultContractResolver
+    {
+        static List<string> PropertiesToIgnore = new List<string>
```
mauroservienti: Since we’re not touching it, is this better as a `string[]` instead of a `List<>`. Nitpicking, probably.

src/NServiceBus.AmazonSQS/Configure/SqsTransportSettings.cs
SeanFeldman: @mauroservienti the intent _is_ to be backwards compatible by default and opt out when that compatibility is no longer needed.

mauroservienti: As far as I can tell this is not what the code does, by default `V1CompatibilityMode` is not set thus it’ll be `false`.
And in https://github.com/Particular/NServiceBus.AmazonSQS/pull/214#issuecomment-386191689 it seems to me we stated that by default v3 is NOT backward compatible, which means that compatibility with v1 needs to be explicitly turned on.

Am I missing something obvious?

boblangley: The other methods had the same default variable value, so I went with it. I am happy to remove it.

src/NServiceBus.AmazonSQS/ReducedPayloadContractResolver.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    using System.Collections.Generic;
+    using System.Reflection;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Serialization;
+    class ReducedPayloadContractResolver : DefaultContractResolver
+    {
+        static IList<string> PropertiesToIgnore = new []
+        {
+            "TimeToBeReceived",
```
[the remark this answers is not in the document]
SeanFeldman: Looks like I've found a spot!
https://assets-auto.rbl.ms/9490c5641a30bfb8c7635cf6c81dac5b9a89c9641912969b1edae950ae33253d

src/NServiceBus.AmazonSQS/MessageDispatcher.cs
```diff
@@ -49,7 +50,14 @@ public MessageDispatcher(ConnectionConfiguration configuration, IAmazonS3 s3Clie
         async Task Dispatch(UnicastTransportOperation unicastMessage)
         {
             var sqsTransportMessage = new TransportMessage(unicastMessage.Message, unicastMessage.DeliveryConstraints);
-            var serializedMessage = JsonConvert.SerializeObject(sqsTransportMessage);
+
+            var jsonSerializerSettings = new JsonSerializerSettings
+            {
```
danielmarbach: Is it necessary to create these settings for each dispatch?

src/NServiceBus.AmazonSQS/ReducedPayloadContractResolver.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+            nameof(TransportMessage.TimeToBeReceived),
+            nameof(TransportMessage.ReplyToAddress)
+        };
+
+        protected override JsonProperty CreateProperty(MemberInfo member, MemberSerialization memberSerialization)
+        {
+            var property = base.CreateProperty(member, memberSerialization);
+
+            property.ShouldSerialize = instance => property.DeclaringType != typeof(TransportMessage) || !PropertiesToIgnore.Contains(property.PropertyName);
```
danielmarbach: What is the impact of the contains vs a direct dict lookup or even better a hardcoded if/else?

boblangley:
Since it looks like I have consensus I am going to close this and adjust this branch to target `release-4.0` and submit under another PR.

PR #215, opened 2018-05-04

boblangley:
Connects to #213

The intent is to:


SeanFeldman:
@boblangley might need to change this to be a a `hotfix-` rather than `release-` to adhear to the PoA. Or change PoA if it's no longer relevant.

PR #216, opened 2018-05-04
V1 Compatible TransportMessage

boblangley:
Connects to #213

The intent is to:


SeanFeldman:
Aside from serializer question @danielmarbach already brought up, I don't have anything else.

In regards to https://github.com/Particular/NServiceBus.AmazonSQS/pull/216#discussion_r186254082 - feels like an ultra-perf optimization. Wherever it's worth it or not, I don't know.

danielmarbach:
The contains lookup is expensive. Considering that the code serves a very specific scenario for back compat which is unlikely to change I see no reason to go with the generic approach that is geared towards extensibility

PR #218, opened 2018-05-11

boblangley:
Connects to #213

The intent is to:


src/NServiceBus.AmazonSQS/TransportConfiguration.cs
```diff
@@ -135,6 +148,7 @@ public int DelayedDeliveryQueueDelayTime
         string queueNamePrefix;
         bool? isDelayedDeliveryEnabled;
         bool? preTruncateQueueNames;
+		bool? useV1CompatiblePayload;
```
SeanFeldman: nit: accidental tab? 😃

src/NServiceBus.AmazonSQS/TransportMessage.cs
```diff
@@ -34,7 +34,7 @@ public TransportMessage(OutgoingMessage outgoingMessage, List<DeliveryConstraint
             Body = outgoingMessage.Body != null ? Convert.ToBase64String(outgoingMessage.Body) : "empty message";
         }
 
-        public Dictionary<string, string> Headers { get; set; } = new Dictionary<string, string>();
+        public Dictionary<string, string> Headers { get; set; }
```
boblangley: Remove the allocation since it is always overwritten in either constructor. The only issue is that property order in serialization matters in native integration scenarios: `Headers` needs to be defined in the JSON before `TimeToBeReceived` or `ReplyToAddress`.

Issue #227, opened 2018-06-25
Improve dispatcher failure message when destination queue does not exist


danielmarbach:
Currently, we only have a global catch for all the dispatches that are fired. Even if just one of the dispatch operation fails the exception is almost meaningless for the end-user. There is no way you can troubleshoot and pin it down to the queue that does not exist based on the following exception message

```
2018-06-22 19:07:28.620 ERROR NServiceBus.Transports.SQS.MessageDispatcher Exception from Send.
Amazon.SQS.Model.QueueDoesNotExistException: The specified queue does not exist for this wsdl version. ---> Amazon.Runtime.Internal.HttpErrorResponseException: The remote server returned an error: (400) Bad Request. ---> System.Net.WebException: The remote server returned an error: (400) Bad Request.
   at System.Net.HttpWebRequest.EndGetResponse(IAsyncResult asyncResult)
```

https://github.com/Particular/NServiceBus.AmazonSQS/blob/support-3.3/src/NServiceBus.AmazonSQS/MessageDispatcher.cs#L50

Further details

https://discuss.particular.net/t/amazon-sqs-queue-does-not-found-with-nsb7/626/4?u=danielmarbach

danielmarbach:
This is done with https://github.com/Particular/NServiceBus.AmazonSQS/pull/229

PR #228, opened 2018-06-25
[WIP] support to log exceptions for each send operations at dispatch time

mauroservienti:
This is a proposed fix for #227. Actually raised as a `hotfix` to `master`, if we don't want to treat this as hotfix we can re-target to `develop` instead.


mauroservienti:
Closing in favor of #229

PR #229, opened 2018-06-26
Better error logging when message sends fail

mauroservienti:
Alternative approach to #228 

This is a proposed fix for #227. Actually raised as a maintenance fix to `develop`, if we want we can treat this as `hotfix` and re-target to `master` instead.

Another approach could be to append a continuation to the `Task` returned by `Dispatch` https://github.com/Particular/NServiceBus.AmazonSQS/blob/efb1869ff9a6a9b1ff30b8fb7dba4a0ad9962af2/src/NServiceBus.AmazonSQS/MessageDispatcher.cs#L42 and log in the continuation. I personally find the approach used in this PR leaner and cleaner.

I vote to close #228 in favor of this one.

### DIFF
--- a/src/NServiceBus.AmazonSQS/MessageDispatcher.cs
+++ b/src/NServiceBus.AmazonSQS/MessageDispatcher.cs
@@ -1,9 +1,5 @@
 ﻿namespace NServiceBus.Transports.SQS
 {
-    using System;
-    using System.IO;
-    using System.Linq;
-    using System.Threading.Tasks;
     using Amazon.S3;
     using Amazon.S3.Model;
     using Amazon.SQS;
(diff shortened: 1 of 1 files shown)


mauroservienti:
Removed WIP

danielmarbach:
We are investigating why the build is failing on Linux. We have a hunch it is related to the .NET Core runtime

danielmarbach:
Marked as WIP until the develop is green again on linux

danielmarbach:
Will be releasing this improvement as part of [4.1.1](https://github.com/Particular/NServiceBus.AmazonSQS/pull/234/commits/2643b4b3f8ffdb1f1ae127331069cb6c52bda51e) which then will be merged to develop

PR #253, opened 2018-11-19
Fix body deletion

danielmarbach:


src/NServiceBus.AmazonSQS/MessagePump.cs
```diff
@@ -438,35 +438,9 @@ async Task DeleteMessageAndBodyIfRequired(Message message, TransportMessage tran
                 return; // if another receiver fetches the data from S3
             }
 
-            if (transportMessage != null)
+            if (!string.IsNullOrEmpty(transportMessage?.S3BodyKey))
```
mauroservienti: should this be `IsNullOrWhiteSpace` just to be sure we're not emitting false positive logs info?

mauroservienti: Fine by me.

PR #273, opened 2019-02-15
Receive messages concurrently without synchronizing

### LINKED ISSUE #242: Task.WhenAll in message pump can block receive operations
[Task.WhenAll](https://github.com/Particular/NServiceBus.AmazonSQS/blob/develop/src/NServiceBus.AmazonSQS/MessagePump.cs#L261) in message pump can block receive operations when going through retries or when one message takes longer than another

danielmarbach:
Closes #242

### DIFF
--- a/src/NServiceBus.AmazonSQS/MessagePump.cs
+++ b/src/NServiceBus.AmazonSQS/MessagePump.cs
@@ -137,8 +137,13 @@
             cancellationTokenSource?.Cancel();
 
             await Task.WhenAll(pumpTasks).ConfigureAwait(false);
-
             pumpTasks?.Clear();
+            
+            while (maxConcurrencySempahore.CurrentCount != maxConcurrency)
+            {
+                await Task.Delay(50).ConfigureAwait(false);
+            }
+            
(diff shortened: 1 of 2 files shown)


src/NServiceBus.AmazonSQS/MessagePump.cs
```diff
@@ -137,8 +137,13 @@ public async Task Stop()
             cancellationTokenSource?.Cancel();
 
             await Task.WhenAll(pumpTasks).ConfigureAwait(false);
-
             pumpTasks?.Clear();
+            
+            while (maxConcurrencySempahore.CurrentCount != maxConcurrency)
```
mauroservienti: if at this point all the pump tasks are gone what are we waiting for?

danielmarbach: Messages in flight that are about to complete. So essentially they passed the semaphore and might already called the core

mauroservienti: got it thanks.

PR #274, opened 2019-02-15
Make NServiceBus aware of the native message id

danielmarbach:
MessageContext and ErrorContext is the interaction point with core

This makes also sure we propagate the native message id into the core that the core will use in case the message id header is not present

### DIFF
--- a/src/NServiceBus.AmazonSQS/MessagePump.cs
+++ b/src/NServiceBus.AmazonSQS/MessagePump.cs
@@ -299,15 +299,16 @@
 
             try
             {
-                IncomingMessage incomingMessage = null;
+                byte[] messageBody = null;
                 TransportMessage transportMessage = null;
+                var messageId = receivedMessage.MessageId;
 
                 var isPoisonMessage = false;
                 try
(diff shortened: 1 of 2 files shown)


src/NServiceBus.AmazonSQS/MessagePump.cs
```diff
@@ -321,7 +322,7 @@ async Task ProcessMessage(Message receivedMessage, CancellationToken token)
                     isPoisonMessage = true;
                 }
 
-                if (incomingMessage == null || transportMessage == null)
+                if (messageBody == null || transportMessage == null)
                 {
                     Logger.Warn($"Treating message with SQS Message Id {receivedMessage.MessageId} as a poison message because it could not be converted to an IncomingMessage. Moving to error queue.");
```
mauroservienti: `IncomingMessage` is not anymore a thing, the log message should be changed.

danielmarbach: My brain read this is incoming message. Good catch

PR #275, opened 2019-02-15
Propagate NSB message ID to SQS message attributes

### LINKED ISSUE #258: Propagate NSB Message ID as SQS Message Attribute
Proposal by @chrissimon-au raised as a follow up in https://github.com/Particular/NServiceBus.AmazonSQS/issues/251#issuecomment-440172425

> Another idea on the behavior of the transport in this area that could have simplified the diagnosis of this issue - although the description above is (relatively) concise, it was quite a winding road to get there:
> 
> The poison message log only includes the SQS Message ID of the poison message.  This is not very helpful, as the copy of the message placed on the next queue (e.g. `error` queue) has a new SQS Message ID. This made it hard for us to isolate the cause of the missing messages as we were searching our logs for the message ID and finding nothing.  It makes sense that it can't, because the poison message cannot be deserialised, so the NSB Message ID cannot be obtained.
> 
> Proposal: add the NSB Message ID as an SQS Message Attribute, so that it can be logged even if the message deserialization fails.

kbaley:
Closes #258 

We decided to propagate the Headers.MessageId value every time. There is no way to opt-in or opt-out from that behavior. We decided against opt-in because we thought in cases of poison messages you'd wish you had it enabled. We decided against opt-out even though you can only use up to 10 message attributes and they contribute to the overall message size because we think the information is always valuable. Plus us reserving one of these 10 message attributes should not be the tipping point at all. If it turns out to be for some reason we can always ship a new minor that allows to opt-out. Right now the additional configuration APIs and code complexity seem unnecessary.

### DIFF
--- a/src/NServiceBus.AmazonSQS/MessagePump.cs
+++ b/src/NServiceBus.AmazonSQS/MessagePump.cs
@@ -321,12 +321,7 @@
                     isPoisonMessage = true;
                 }
 
-                if (messageBody == null || transportMessage == null)
-                {
-                    isPoisonMessage = true;
-                }
-
-                if (isPoisonMessage)
+                if (isPoisonMessage || messageBody == null || transportMessage == null)
                 {
                     var logMessage = $"Treating message with {messageId} as a poison message. Moving to error queue.";
                     
--- a/src/NServiceBus.AmazonSQS/MessagePump.cs
+++ b/src/NServiceBus.AmazonSQS/MessagePump.cs
@@ -103,10 +103,8 @@
                 MaxNumberOfMessages = numberOfMessagesToFetch,
                 QueueUrl = queueUrl,
                 WaitTimeSeconds = 20,
-                AttributeNames = new List<string>
-                {
-                    "SentTimestamp"
-                }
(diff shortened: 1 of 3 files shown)


src/NServiceBus.AmazonSQS/MessagePump.cs
```diff
@@ -308,39 +317,50 @@ async Task ProcessMessage(Message receivedMessage, CancellationToken token)
                 catch (Exception ex)
                 {
                     // Can't deserialize. This is a poison message
-                    Logger.Warn($"Treating message with SQS Message Id {receivedMessage.MessageId} as a poison message due to exception {ex}. Moving to error queue.");
+                    exception = ex;
                     isPoisonMessage = true;
                 }
 
                 if (messageBody == null || transportMessage == null)
```
mauroservienti: can't this be merged into the next condition as:

```
if (isPoisonMessage || messageBody == null || transportMessage == null)
```
```

danielmarbach: Good catch. Fixed

PR #276, opened 2019-02-15
Add better error message for client creation

### LINKED ISSUE #187: Error messages when configuration is bad are not helpful
e.g. when SQS is configured properly but S3 is not user gets an exception that does not mention where the problem is like "Region or ServiceURL is not configured"

danielmarbach:
Closes #187

Technically this is a behavior breaking change because we wrap the exception that is raised with an outer exception that contains a more meaningful exception message and the original exception as an inner exception. We decided against bumping the major because of the following reasons:

- Already today the SDK behaves differently between .NET Full framework and .NET Core. While the full framework raises an AmazonClientException the .NET Core version raises a `HttpResponseException`, more information available at https://github.com/aws/aws-sdk-net/issues/1216. Afaik the behavior is already today different depending on the framework that is running the code
- The exception that is raised is intransparent and doesn't lead users to the pit of success. For example the full framework version could raise (depending on the configuration error) `Unhandled Exception: Amazon.Runtime.AmazonClientException: No RegionEndpoint or ServiceURL configured` while the .NET Core version would raise `Unhandled Exception: System.NET.HttpRequestException: Connection refused...`
- The previous exception happened on a bootstrapping path. It is probably valid to assume that existing customers already have that path properly configured. In case their code would assume an `AmazonClientException` to be thrown the code would break when started and it would be noticed during deployment to one of the lower environments in the worst case when the production deployment happens. In the latter case, the problem can easily be fixed by switching the exception type caught. Another potentially valid assumption is that in the bootstrapping code customers might already have more generous try-catch blocks that would at least properly log what has happened. 

Taking into account the above reasons in combination that due to missconfiguration customers could spend hours fiddling around we decided to include these changes in a minor and be a bit more relaxed about SemVer.

### DIFF
--- a/src/NServiceBus.AmazonSQS/Configure/TransportInfrastructure.cs
+++ b/src/NServiceBus.AmazonSQS/Configure/TransportInfrastructure.cs
@@ -8,6 +8,7 @@
     using Amazon.SQS;
     using AmazonSQS;
     using DelayedDelivery;
+    using Logging;
     using Performance.TimeToBeReceived;
     using Routing;
     using Settings;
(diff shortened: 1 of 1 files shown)


src/NServiceBus.AmazonSQS/Configure/TransportInfrastructure.cs
```diff
@@ -20,8 +21,27 @@ public TransportInfrastructure(ReadOnlySettings settings)
         {
             configuration = new TransportConfiguration(settings);
 
-            sqsClient = configuration.SqsClientFactory();
-            s3Client = configuration.S3ClientFactory();
+            try
+            {
+                sqsClient = configuration.SqsClientFactory();
+            }
+            catch (Exception e) // ideally this would just raise AmazonClientException but the fullframework and .net core version behave different, see https://github.com/aws/aws-sdk-net/issues/1216
```
danielmarbach: @mauroservienti I had to update the PR. @kbaley and I found a different behavior between the full framework and .NET Core version of it, more context available in https://github.com/aws/aws-sdk-net/issues/1216

danielmarbach:
Rebased

src/NServiceBus.AmazonSQS/Configure/TransportInfrastructure.cs
mauroservienti: Wouldn’t it be better to catch different exceptions using compiler directives? By catching the broader Exception can’t we run the risk of hiding different kind of errors under the missing credentials umbrella?

danielmarbach: If we did that then we would have to include a package reference in the .NET core specific part to System.Net.Http.

I know that this catch is broad but any exception when creating the client hints either at credential problems or like the generic phrase

> or the client factory configures the created client accordingly

hints a broader problem with client configuration. The inner exception is still logged so while still have the generic catch the log would contain more information to troubleshoot in case we'd run into a more broader problem and the endresult would still be that the endpoint cannot start.

PR #277, opened 2019-02-18
Avoid header manipulation being visible between processing attempts

danielmarbach:
Closes #248, #249, #250 

The core seam does not prevent header manipulations from being leaked back into the transport. The mutable dictionary exposed on the message context practically invites for that problem to happen. Instead of just creating a new dictionary every time a simple fixed size dictionary pool is used that grows on demand. This is more efficient as shown in https://github.com/Particular/NServiceBus.AmazonSQS/issues/248#issuecomment-463616074

The pool grows its capacity over time up to the concurrency of the pump. The pool is not shared between the pumps although technically it could be

This exposes us to the following risk

If someone assigns the header dictionary of an incoming message to a state holder that outlives the message handling pipeline that dictionary will get cleared ones the message is processed by the transport. Users doing that have to copy the headers. This is an acceptable design constraints given that you cannot assume that message based information is going to stay longer than the pipeline (afaik the core even enforces that by floating the state into the handler and pipeline methods.

I verified that with @SzymonPobiega wearing both his "knowledgeable in core" and "keeping the lights on" hat and we agree that it is an acceptable tradeoff because paying for the dictionary allocation every time is not worth it.

Should someone buffer outgoing messages and send them out later those users would not be affected because outgoing messages have dedicated header copies (exception when forwarding but that happens as part of the incoming message handling pipeline)

Issue #250, opened 2018-11-05
Header manipulations done while processing a message are visible in recoverability policy


SzymonPobiega:
When a header is added/deleted/modified in an attempt to process a message and that attempt fails, these header manipulations are visible when the recoverability policy is executed to determine if/how the message should be retried.

danielmarbach:
Addressed with #279
Closing

Issue #249, opened 2018-11-05
Header manipulations done in recoverability policy are visible during next attempt to process a message


SzymonPobiega:
When a header is added/deleted/modified in an attempt to process a message and that attempt fails, these header manipulations are visible when the recoverability policy is executed to determine if/how the message should be retried.

danielmarbach:
Addressed with #279
Closing

PR #279, opened 2019-02-18
Copy headers

### LINKED ISSUE #248: Header manipulations persist between processing attempts
When a header is added/deleted/modified in the first attempt to process a message and that attempt fails, these header manipulations are visible when the message is retried in a subsequent attempt.

danielmarbach:
Closes #248, #249, #250 

Alternative to #277


danielmarbach:
@mauroservienti I can't argue with the explicitness. #277 is more complex because it also addresses the problem of unnecessarily creating tons of dictionaries that's why it is more complex. I'm happy to merge this one as a fix for now and discuss the ramifications of #277 a bit more

PR #281, opened 2019-02-19
Batching

### LINKED ISSUE #102: Dispatcher: Use batched sends for multiple messages as long as payload fits size restrictions
Like ASB the SQS transport could try to estimate the payload sized then bucket them into multiple buckets of N number of messages that still fit the size restrictions (256 KB) and send them in batches for non isolated dispatches

danielmarbach:
Closes #102

### DIFF
--- a/src/AcceptanceTests/Batching/When_batching_multiple_outgoing_small_messages.cs
+++ b/src/AcceptanceTests/Batching/When_batching_multiple_outgoing_small_messages.cs
@@ -133,7 +133,7 @@ namespace NServiceBus.AcceptanceTests.Batching
         {
             public Receiver()
             {
-                EndpointSetup<DefaultServer>();
+                EndpointSetup<DefaultServer>(c => c.LimitMessageProcessingConcurrencyTo(60));
             }
 
             public class MyMessageHandler : IHandleMessages<MyMessage>
--- a/src/AcceptanceTests/Batching/When_batching_multiple_outgoing_small_messages.cs
+++ b/src/AcceptanceTests/Batching/When_batching_multiple_outgoing_small_messages.cs
@@ -1,7 +1,10 @@
 namespace NServiceBus.AcceptanceTests.Batching
 {
     using System;
-    using System.Threading;
+    using System.Collections.Concurrent;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Text;
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using AcceptanceTesting.Customization;
@@ -16,7 +19,27 @@ namespace NServiceBus.AcceptanceTests.Batching
         {
             var kickOffMessageId = Guid.NewGuid().ToString();
 
-            var context = await Scenario.Define<Context>(c => c.LogLevel = LogLevel.Debug)
+            var listOfMessagesForBatching = new List<string>();
+            for (var i = 0; i < 50; i++)
+            {
+                listOfMessagesForBatching.Add(Guid.NewGuid().ToString());
+            }
+
+            var listOfMessagesForImmediateDispatch = new List<string>();
+            for (var i = 0; i < 10; i++)
+            {
+                listOfMessagesForImmediateDispatch.Add(Guid.NewGuid().ToString());
+            }
+
+            var allMessageids = listOfMessagesForBatching.Union(listOfMessagesForImmediateDispatch).ToList();
+
+            var context = await Scenario.Define<Context>(c =>
+                {
+
+                    c.LogLevel = LogLevel.Debug;
+                    c.MessageIdsForBatching = listOfMessagesForBatching;
+                    c.MessageIdsForImmediateDispatch = listOfMessagesForImmediateDispatch;
+                })
                 .WithEndpoint<Sender>(b => b.When(session =>
                 {
                     var options = new SendOptions();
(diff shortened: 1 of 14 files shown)


danielmarbach:
So this is ready for review and intensive testing

src/NServiceBus.AmazonSQS/Batcher.cs
```diff
@@ -0,0 +1,57 @@
...
+                    firstMessage = firstMessage ?? message;
+
+                    var bodyLength = message.Body?.Length ?? 0;
+                    payloadSize += bodyLength;
+
+                    if (payloadSize > TransportConfiguration.MaximumMessageSize)
+                    {
+                        allBatches.Add(message.ToBatchRequest(currentDestinationBatches));
+                        currentDestinationBatches.Clear();
+                        payloadSize = bodyLength;
```
kbaley: Why is payloadSize reset to the size of the current message body instead of subtracting `bodyLength` from the current value?

src/NServiceBus.AmazonSQS/Batcher.cs
```diff
@@ -0,0 +1,57 @@
...
+                    }
+
+                    // we don't have to recheck payload size here because the upport layer checks that a request can always fit 256 KB size limit
```
kbaley:
```suggestion
                    // we don't have to recheck payload size here because the support layer checks that a request can always fit 256 KB size limit
```

src/AcceptanceTests/Batching/When_batching_multiple_outgoing_small_messages.cs
```diff
@@ -0,0 +1,159 @@
...
+                StringAssert.Contains(messageIdForBatching, logoutput, $"{messageIdForBatching} not found in any of the batches. Output: {logoutput}");
+            }
+
+            foreach (var messageIdForImmediateDispatch in listOfMessagesForImmediateDispatch)
+            {
+                StringAssert.DoesNotContain(messageIdForImmediateDispatch, logoutput, $"{messageIdForImmediateDispatch} was found in any of the batches. Output: {logoutput}");
+            }
+
+            // let's see how many times this actually happens
+            StringAssert.DoesNotContain("Retried message with MessageId", "Messages have been retried but they shouldn't have been");
```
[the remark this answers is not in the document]
danielmarbach: Will fix

src/NServiceBus.AmazonSQS/Batcher.cs
danielmarbach: When a batch is prepared either due to messagd size or number of messages in batch we can restart the payload calculations

src/NServiceBus.AmazonSQS/PreparedMessage.cs
```diff
@@ -0,0 +1,18 @@
...
+    using Amazon.SQS.Model;
+
+    class PreparedMessage
+    {
+        public string MessageId { get; set; }
+        public string Body{ get; set; }
+        public string Destination { get; set; }
+        public string OriginalDestination { get; set; }
+        public string QueueUrl { get; set; }
+        public long DelaySeconds { get; set; }
```
mauroservienti: For my curiosity what is the point of having `DelaySeconds` being a `long` if when used it need to be converted to an `int` anyway?

src/NServiceBus.AmazonSQS/Batcher.cs
```diff
@@ -0,0 +1,57 @@
...
+    using AmazonSQS;
+
+    static class Batcher
+    {
+        public static IReadOnlyList<BatchEntry> Batch(IEnumerable<PreparedMessage> preparedMessages)
+        {
+            var allBatches = new List<BatchEntry>();
+            var currentDestinationBatches = new Dictionary<string, PreparedMessage>();
+
+            var groupByDestination = preparedMessages.GroupBy(m => m.QueueUrl, StringComparer.OrdinalIgnoreCase);
```
mauroservienti: given that SQS queue names and URLs are case-sensitive (IIRC) does this have the risk of putting in the same batch messages for different destinations? e.g. `Sales` and `sales` will be in the same group.

src/NServiceBus.AmazonSQS/Batcher.cs
```diff
@@ -0,0 +1,57 @@
...
+            foreach (var group in groupByDestination)
+            {
+                PreparedMessage firstMessage = null;
+                var payloadSize = 0;
+                for (var i = 0; i < group.Count(); i++)
```
mauroservienti: nitpicking/curiosity: given that `Count()` enumerates all items to retrieve the count, and then a for is performed, isn't a `foreach` a better alternative?

src/NServiceBus.AmazonSQS/MessageDispatcher.cs
```diff
@@ -33,25 +34,167 @@ public MessageDispatcher(TransportConfiguration configuration, IAmazonS3 s3Clien
...
+
+            return tasks != null ? Task.WhenAll(tasks) : TaskExtensions.Completed;
+        }
 
+        async Task DispatchBatched(IEnumerable<UnicastTransportOperation> toBeBatchedTransportOperations)
+        {
+            List<Task<PreparedMessage>> tasks = null;
+            foreach (var operation in toBeBatchedTransportOperations)
+            {
+                tasks = tasks ?? new List<Task<PreparedMessage>>();
```
mauroservienti: For my curiosity: is this only to avoid an allocation? Is it cheaper to constantly perform the `if` branching?

src/NServiceBus.AmazonSQS/MessageDispatcher.cs
```diff
@@ -33,25 +34,167 @@ public MessageDispatcher(TransportConfiguration configuration, IAmazonS3 s3Clien
...
+                tasks.Add(PrepareMessage(operation));
+            }
+
+            if (tasks != null)
```
mauroservienti: Also: I cannot spot a path in which `DispatchBatched` can be called with an empty `IEnumerable<UnicastTransportOperation>`. If called there will be messages to dispatch, previously in the code:

```
 case DispatchConsistency.Default:
   concurrentDispatchTasks = concurrentDispatchTasks ?? new List<Task>();
   concurrentDispatchTasks.Add(DispatchBatched(dispatchConsistencyGroup));
   break;
```

Am I missing something?

src/NServiceBus.AmazonSQS/Batcher.cs
danielmarbach: Great catch. I wanted to use Ordinal only and not ignorecase

src/NServiceBus.AmazonSQS/Batcher.cs
danielmarbach: Yeah will switch

src/NServiceBus.AmazonSQS/MessageDispatcher.cs
danielmarbach: You are right I can eager alloc

src/NServiceBus.AmazonSQS/PreparedMessage.cs
danielmarbach: I think I left it a long because it was a long before. Lemme check

src/NServiceBus.AmazonSQS/MessageDispatcher.cs
danielmarbach: Yes because if we have nothing to enumerate we don't allocate a list. If else should be cheaper than paying for GC

src/NServiceBus.AmazonSQS/PreparedMessage.cs
danielmarbach: So for the calculation, we need it to be a long because we support with UnrestrictedDelayDelivery arbitrary long intervals. But given that prepared message is used to turn it into DelaySeconds of the type this can be an int and the conversation can happen when creating the prepared message

danielmarbach:
@mauroservienti addressed your comments

danielmarbach:
@kbaley @mauroservienti once approved I will merge it

PR #283, opened 2019-02-20
Migrate to SimpleJson

### LINKED ISSUE #189: Should we use SimpleJson for the transport envelope serialization and deserialization
This would allow us to get rid of the newtonsoft json dependency.

We would need to check the backward compliance from SimpleJson and NewtonsoftJson but it looks like it should be possible because the envelope is a really simple and lightweight data type.

danielmarbach:
Closes #189


src/NServiceBus.AmazonSQS/MessagePump.cs
```diff
@@ -305,8 +305,7 @@ async Task ProcessMessage(Message receivedMessage, CancellationToken token)
                         messageId = nativeMessageId;
                     }
 
-                    transportMessage = JsonConvert.DeserializeObject<TransportMessage>(receivedMessage.Body);
-
+                    transportMessage = SimpleJson.DeserializeObject<TransportMessage>(receivedMessage.Body);
```
danielmarbach: we don't need to use the strategy here since we only care about V1 backwards compat serialization for outgoing messages

src/NServiceBus.AmazonSQS/NServiceBus.AmazonSQS.csproj
```diff
@@ -9,7 +9,6 @@
     <PackageReference Include="AWSSDK.S3" Version="[3.3.18.5, 3.4)" />
     <PackageReference Include="AWSSDK.SQS" Version="[3.3.3.10, 3.4)" />
     <PackageReference Include="NServiceBus" Version="[7.0.0, 8.0.0)" />
-    <PackageReference Include="Newtonsoft.Json" Version="[12.0.1, 13.0.0)" />
```
danielmarbach: 🎆

danielmarbach:
LGTM

kbaley:
I smoke tested this locally sending messages with JSON.NET and receiving them with SimpleJson, and vice versa and all went well. Just need to address the failing test.

danielmarbach:
Sent in an alternative PR that doesn't require us to switch away from the struct https://github.com/Particular/NServiceBus.AmazonSQS/pull/286

PR #286, opened 2019-02-21
Simplejson blazor

danielmarbach:
Alternative to #283 

Compatibility mode to v5 receiver

Left with Newtonsoft Json, right with SimpleJson

image


src/NServiceBus.AmazonSQS/SimpleJson/SimpleJson.g.cs
```diff
@@ -0,0 +1,2156 @@
...
+//    limitations under the License.
+// </copyright>
+// <author>Nathan Totten (ntotten.com), Jim Zimmerman (jimzimmerman.com) and Prabir Shrestha (prabir.me)</author>
+// <website>https://github.com/facebook-csharp-sdk/simple-json</website>
+//-----------------------------------------------------------------------
+
+// VERSION:
+
+// NOTE: uncomment the following line to make SimpleJson class internal.
+#define SIMPLE_JSON_INTERNAL
```
boblangley: Should we make this all internal? Is there a reason we should make this available outside of the transport?

danielmarbach: It is internal because the define is uncommented

boblangley: Are you sure, am I reading this wrong?

https://github.com/Particular/NServiceBus.AmazonSQS/blob/3271c36541bd7c4d856f0e39169f5e376f029047/src/NServiceBus.AmazonSQS/SimpleJson/SimpleJson.g.cs#L497-L502

boblangley: Nevermind. I am reading it wrong.

PR #290, opened 2019-04-02
Use MemoryStream instead of BufferedStream

mattcanty:
S3 response stream does not support seeking therefore calling `.Length` fails. Also the logic to create the byte array does not work. Using `MemoryStream.ToArray()` is sufficient.

### DIFF
--- a/src/NServiceBus.AmazonSQS/Extensions/MessageExtensions.cs
+++ b/src/NServiceBus.AmazonSQS/Extensions/MessageExtensions.cs
@@ -26,18 +26,10 @@
                     transportMessage.S3BodyKey,
                     cancellationToken).ConfigureAwait(false);
 
-                body = new byte[s3GetResponse.ResponseStream.Length];
-                using (var bufferedStream = new BufferedStream(s3GetResponse.ResponseStream))
+                using (var memoryStream = new MemoryStream())
                 {
-                    int count;
-                    var transferred = 0;
-                    const int maxChunkSize = 8 * 1024;
-                    var bytesToRead = Math.Min(maxChunkSize, body.Length - transferred);
-                    while ((count = await bufferedStream.ReadAsync(body, transferred, bytesToRead, cancellationToken).ConfigureAwait(false)) > 0)
-                    {
-                        transferred += count;
-                        bytesToRead = Math.Min(maxChunkSize, body.Length - transferred);
-                    }
+                    s3GetResponse.ResponseStream.CopyTo(memoryStream);
+                    body = memoryStream.ToArray();
                 }
             }
 


danielmarbach:
Hi Matt,

Thanks for submitting a PR. Can you tell us when this stopped working? Which SDK version are you using? This code has been used in various versions of the transport and was written by the original maintainer. It also passes our acceptance tests which use S3 as storage so I'm curious why is not working on your end

mattcanty:
@danielmarbach still building up a picture but it appears to be a problem with various versions. We know that this is currently affecting version 3.6.2 of ServiceControl, however we believe there are other issues with that too.

## To reproduce
Create a console application, output below
```
using System;
using System.IO;
using System.Text;
using System.Threading.Tasks;
using Amazon;
using Amazon.S3;
using Amazon.SecurityToken;
using Amazon.SecurityToken.Model;

namespace S3Test
{
    class MainClass
    {
        static readonly string awsAccessKeyId = "REDACTED";
        static readonly string awsSecretAccessKey = "REDACTED";
        static readonly string mfaSerialNumber = "REDACTED";
        static readonly string roleArn = "REDACTED";

        static readonly string bucketName = "REDACTED";
        static readonly string key = "REDACTED";

        public static void Main(string[] args)
        {
            Console.Write("Enter MFA code: ");
            var mfaTOTP = Console.ReadLine();

            var stsClient = new AmazonSecurityTokenServiceClient(awsAccessKeyId, awsSecretAccessKey);

            var credentials = stsClient.AssumeRole(new AssumeRoleRequest
            {
                RoleArn = roleArn,
                RoleSessionName = "dotnettest",
                DurationSeconds = 3600,
                SerialNumber = mfaSerialNumber,
                TokenCode = mfaTOTP
            }).Credentials;

            var regionEndpoint = RegionEndpoint.EUWest1;
            var amazonS3 = new AmazonS3Client(credentials, regionEndpoint);

            Task.Run(async () =>
            {
                try
                {
                    Console.WriteLine("Method A");

                    var bytes = await RetrieveBodyA(amazonS3);

                    Console.WriteLine(Encoding.Default.GetString(bytes).Substring(0, 50));
                }
                catch (NotSupportedException e)
                {
                    Console.WriteLine($"'Method A' Exception: {e.Message}");
                }

                try
                {
                    Console.WriteLine("Method B");

                    var bytes = await RetrieveBodyB(amazonS3);

                    Console.WriteLine(Encoding.Default.GetString(bytes).Substring(0, 50));
                }
                catch (NotSupportedException e)
                {
                    Console.WriteLine($"'Method B' Exception: {e.Message}");
                }
            }).Wait();
        }

        static async Task<byte[]> RetrieveBodyA(IAmazonS3 amazonS3)
        {
            byte[] body;
            var s3GetResponse = await amazonS3.GetObjectAsync(bucketName, key).ConfigureAwait(false);

            body = new byte[s3GetResponse.ResponseStream.Length];

            using (var bufferedStream = new BufferedStream(s3GetResponse.ResponseStream))
            {
                int count;
                var transferred = 0;
                const int maxChunkSize = 8 * 1024;
                var bytesToRead = Math.Min(maxChunkSize, body.Length - transferred);
                while ((count = await bufferedStream.ReadAsync(body, transferred, bytesToRead).ConfigureAwait(false)) > 0)
                {
                    transferred += count;
                    bytesToRead = Math.Min(maxChunkSize, body.Length - transferred);
                }
            }

            return body;
        }

        static async Task<byte[]> RetrieveBodyB(IAmazonS3 amazonS3)
        {
            var s3GetResponse = await amazonS3.GetObjectAsync(bucketName, key).ConfigureAwait(false);

            using (var memoryStream = new MemoryStream())
            {
                s3GetResponse.ResponseStream.CopyTo(memoryStream);
                return memoryStream.ToArray();
            }
        }
    }
}
```

Here is the console output:
```
Enter MFA code: 035234
Method A
'Method A' Exception: This stream does not support seek operations.
Method B
{"RefreshCorrelationId":"3277f232-3bfb-4dee-b529-

Press any key to continue...
```

mattcanty:
I've checked through your acceptance tests. It doesn't look to me like the method is covered. I've just searched for the method usages via Git's web interface so apologies if I've missed them.

It looks like where the tests should be does not call the extension method: https://github.com/Particular/NServiceBus.AmazonSQS/blob/277cb8a67b151435c40059f5065088902afdc406/src/AcceptanceTests/Sending_messages_with_large_payloads.cs

danielmarbach:
It is here https://github.com/Particular/NServiceBus.AmazonSQS/blob/develop/src/AcceptanceTests/SqsTransportExtensions.cs#L23

I executed it again locally on my machine using

- NET452
- NETCOREAPP2.1
- NETCOREAPP2.2

I also bumped the dependencies to latest S3 and SQS SDK, namely

```
    <PackageReference Include="AWSSDK.S3" Version="3.3.101.1" />
    <PackageReference Include="AWSSDK.SQS" Version="3.3.100.3" />
```

mattcanty:
I don't believe the extension method is ever called. The type `Amazon.S3.Model.StreamResponse` does not support `seek`. I added log lines to the above reproducible code to show that `MemoryStream` does support `seek`.

danielmarbach:
It is called by our acceptance test infrastructure

see

image

image


Would you mind uploading your csproj to a github repo and add the repo link here?

Does this happen under Windows or linux only? Does this happen only under full framework or only the netcoreapp?

mattcanty:
I've uploaded my solution - https://github.com/mattcanty/S3Test

mattcanty:
Observed issue in:
* 4.6.2
* 4.7

danielmarbach:
@mattcanty I digged deeper today and looked into the AWS SDK. When I did that I realized that the underlying stream changes depending on the S3 encryption policy that is used on the object. So the buffered stream approach and reading the length property works for

- NONE
- AES-256

image

but it fails for 

- AWS-KMS

because then the stream is wrapped with `DecryptStream` which doesn't support seeking

Here the output of your repro solution augment a bit

### NONE

```
Method A
Amazon.S3.Model.StreamResponse Supports Seek: False
Amazon.S3.Model.StreamResponse Supports Length: 205049
Amazon.S3.Model.StreamResponse Stream Type: Amazon.Runtime.Internal.Util.MD5Stream
<?xml version="1.0"?><MyMessageWithLargePayload xm
Method B
System.IO.MemoryStream Supports Seek: True
System.IO.MemoryStream Supports Seek: 0
<?xml version="1.0"?><MyMessageWithLargePayload xm
```

### AES-256

```
Method A
Amazon.S3.Model.StreamResponse Supports Seek: False
Amazon.S3.Model.StreamResponse Supports Length: 205049
Amazon.S3.Model.StreamResponse Stream Type: Amazon.Runtime.Internal.Util.MD5Stream
<?xml version="1.0"?><MyMessageWithLargePayload xm
Method B
System.IO.MemoryStream Supports Seek: True
System.IO.MemoryStream Supports Seek: 0
<?xml version="1.0"?><MyMessageWithLargePayload xm
```

### AWS-KMS

```
Method A
Amazon.S3.Model.StreamResponse Supports Seek: False
'Method A' Exception: This stream does not support seek operations.
Method B
System.IO.MemoryStream Supports Seek: True
System.IO.MemoryStream Supports Seek: 0
<?xml version="1.0"?><MyMessageWithLargePayload xm
```

Can you confirm that you have AWS-KMS enabled on the S3 bucket you are testing against?

mattcanty:
@danielmarbach yes we are using AWS-KMS. Good find!

danielmarbach:
Do you have a hard requirement for AWS-KMS for S3 due to some regulations or would it be OK for you to use AES-256 only for the time until we solved this issue? I'm aware that AWS-KMS has more stringent security constraints on accessing the objects and I'm sure you want to prefer that but I'm asking to see how much actual priority this has for you and your company

mattcanty:
We're going to assess it with our security team, but I would hazard a guess that KMS would definitely be preferred.

danielmarbach:
Raised https://github.com/Particular/NServiceBus.AmazonSQS/issues/291

Will also raise an internal triage issue and update you with the outcome as soon as I know more

mattcanty:
We're going to investigate using AES today.

danielmarbach:
It would be really temporary until we fixed it. Currently we are looking at the ServiceControl issue and this one that you raised and are consider it a priority. Which means we'll start addressing it as soon as the other priority issues that are current work in progress will be fixed and we are likely to bundle those two things to get you unstuck. I'll keep you posted

mattcanty:
Feedback from our team is that we’d prefer a priority on this issue as we can solve ServiceControl temporarily ourselves quite easily.

As soon as you can provide a time estimate that would be great as it would be a biggish job to switch out the bucket encryption type.

mattcanty:
Hi @danielmarbach, has there been any progress with this? Apologies for pinging you on multiple issues today :-)

andreasohlund:
Hi @mattcanty, thanks for keeping in touch. We will start to address this issue tomorrow so stay tuned.

danielmarbach:
@mattcanty could you please sign https://particular.net/contributors-license-agreement-consent ? Thanks

andreasohlund:
We had some issues with building external PR's on our buildserver so we raised https://github.com/Particular/NServiceBus.AmazonSQS/pull/297 instead with a similar change. Are you ok with not having your commit be part of the fix?

danielmarbach:
@mattcanty Manually added your commit to https://github.com/Particular/NServiceBus.AmazonSQS/pull/298

Thanks for the contribution! Can you send an email with your shipping address and with your phone number to daniel dot marbach @ particular dot net . We'd like to send you a small contribution gift.

PR #292, opened 2019-04-12
Hotfix 4.3.1

andreasohlund:
Making sure that we invoke critical error if recoverability policy fails.


andreasohlund:
@boblangley @SeanFeldman this one is ready to go

PR #293, opened 2019-04-15
Hotfix 3.3.4

andreasohlund:
Invoke critical error if recoverability fails

### DIFF
--- a/src/NServiceBus.AmazonSQS/NServiceBus.AmazonSQS.csproj
+++ b/src/NServiceBus.AmazonSQS/NServiceBus.AmazonSQS.csproj
@@ -5,7 +5,7 @@
     <AssemblyName>NServiceBus.AmazonSQS</AssemblyName>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>$(SolutionDir)NServiceBus.snk</AssemblyOriginatorKeyFile>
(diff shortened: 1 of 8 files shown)


src/NServiceBus.AmazonSQS/NServiceBus.AmazonSQS.csproj
```diff
@@ -5,7 +5,7 @@
     <AssemblyName>NServiceBus.AmazonSQS</AssemblyName>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>$(SolutionDir)NServiceBus.snk</AssemblyOriginatorKeyFile>
-    <NoWarn>CS0419</NoWarn>
+    <NoWarn>CS0419;NU5125</NoWarn>
```
andreasohlund: @bording I thought this shouldn't be needed after you going through all the repos? (What am I doing wrong, builds fine locally btw)

bording: It should have been `<NoWarn>$(NoWarn);CS0419</NoWarn>` to not overwrite values being set elsewhere (from Particular.Packaging, for example), but it doesn't look like it's needed at all, so I've just removed it entirely instead.

Issue #291, opened 2019-04-03
Large message bodies encrypted with AWS-KMS cannot be received


danielmarbach:
## Who's affected

- Customers sending large message bodies and using AWS-KMS to encrypt the S3 objects enabled on the bucket

![image](https://user-images.githubusercontent.com/174258/55468261-4c822900-5603-11e9-8418-c76bb5304840.png)

## Symptoms

Messages with large message bodies are sent to S3 but when the transport tries to receive them the receive fails with

```
2019-04-01 13:49:38.6062|31|Warn|NServiceBus.Transports.SQS.MessagePump|Treating message with SQS Message Id 456bd3cd-3720-408a-a2de-245c21e91e7d as a poison message due to exception System.NotSupportedException: This stream does not support seek operations.
   at System.Net.ConnectStream.get_Length()
   at NServiceBus.AmazonSQS.MessageExtensions.<ToIncomingMessage>d__0.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NServiceBus.Transports.SQS.MessagePump.<ProcessMessage>d__7.MoveNext(). Moving to error queue.
```

and messages are moved to the error queue but cannot be retried because the transport will never be able to receive those messages

## Workaround

Switching encryption to AES-256 only allows the messages to be received but has security implications because AWS-KMS offers more stringent security constraints on accessing the objects, including mandatory use of HTTPS and Signature Version 4.

## Analysis

See https://github.com/Particular/NServiceBus.AmazonSQS/pull/290

PR #297, opened 2019-05-01
Switch from BufferedStream to MemoryStream for S3 messages

danielmarbach:
Fixes https://github.com/Particular/NServiceBus.AmazonSQS/issues/291


danielmarbach:
First thought about implementing a dual way to preserve the BufferedStream operation for response streams with `CanSeek` equals `true`. After doing a benchmark and comparing different ressources about buffered stream I came to the conclusion always going with the memorystream for now is good enough

``` ini

BenchmarkDotNet=v0.11.3, OS=Windows 10.0.17763.437 (1809/October2018Update/Redstone5)
AMD Ryzen Threadripper 1920X, 1 CPU, 24 logical and 12 physical cores
.NET Core SDK=2.2.104
  [Host]     : .NET Core 2.1.9 (CoreCLR 4.6.27414.06, CoreFX 4.6.27415.01), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.9 (CoreCLR 4.6.27414.06, CoreFX 4.6.27415.01), 64bit RyuJIT


```
|         Method |     Mean |     Error |   StdDev |    StdErr |      Min |       Q1 |   Median |       Q3 |      Max |  Op/s | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|--------------- |---------:|----------:|---------:|----------:|---------:|---------:|---------:|---------:|---------:|------:|------------:|------------:|------------:|--------------------:|
| BufferedStream | 24.67 ms | 0.6196 ms | 1.807 ms | 0.1826 ms | 21.48 ms | 23.22 ms | 24.37 ms | 25.85 ms | 29.81 ms | 40.53 |     62.5000 |           - |           - |            67.47 KB |
|   MemoryStream | 25.35 ms | 0.6436 ms | 1.857 ms | 0.1895 ms | 21.82 ms | 24.06 ms | 25.10 ms | 26.47 ms | 30.09 ms | 39.46 |     62.5000 |           - |           - |            67.46 KB |

danielmarbach:
I tried to use https://github.com/Microsoft/Microsoft.IO.RecyclableMemoryStream in another spike to rent buffers but could not get it to work because ArraySegment slicing is only supported in Netstandard 2.1

danielmarbach:
We could switch the default bucket to use KMS and then it would be covered with the existing S3 bucket test. Thoughts?

danielmarbach:
Otherwise we would need to setup a dedicated bucket with a the KMS encryption policy and have a dedicated test for that that uses this bucket

andreasohlund:
I guess one test for each bucket type would give most coverage but since this is a hotfix perhaps swtching the default is good enough?(and consider a separate pr against develop)

PR #307, opened 2019-06-10
Use CorrectedUtcNow to handle TTBR

### LINKED ISSUE #210: TTBR implementation does not account for clock skew
Unrestricted Delayed Delivery implementation has accounted for clock skew.
TTBR implementation does not do that and is prone to clock skew.

mauroservienti:
Fix #210

### DIFF
--- a/src/NServiceBus.AmazonSQS/MessagePump.cs
+++ b/src/NServiceBus.AmazonSQS/MessagePump.cs
@@ -7,7 +7,6 @@
     using Amazon.S3;
     using Amazon.SQS;
     using Amazon.SQS.Model;
-    using Amazon.Util;
     using AmazonSQS;
     using Extensibility;
     using Logging;
--- a/src/NServiceBus.AmazonSQS/MessagePump.cs
+++ b/src/NServiceBus.AmazonSQS/MessagePump.cs
@@ -170,7 +170,7 @@
 
                         if (Convert.ToInt32(receivedMessage.Attributes["ApproximateReceiveCount"]) > 1)
                         {
-                            received = DateTimeOffset.UtcNow;
+                            received = sqsClient.Config.CorrectedUtcNow;
                         }
 
                         var elapsed = received - sent;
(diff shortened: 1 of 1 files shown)


danielmarbach:
We also have [DateTime.UtcNow in the dispatcher](https://github.com/Particular/NServiceBus.AmazonSQS/blob/83cdb342b49ab00c42de9b0063fc390565365429/src/NServiceBus.AmazonSQS/MessageDispatcher.cs#L198). I do think though for dispatching it might make sense to leave the client time or should we align?

It seems like the `CorrectedUtcNow` is obsoleted in some versions of the SDK, see https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Core/Amazon.Runtime/ClientConfig.cs#L485 and therefore https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Core/Amazon.Runtime/CorrectClockSkew.cs#L100 should be used?

mauroservienti:
I'm not sure we need to change it now, especially because if we change the way we retrieve information about clock skew we probably want to change it everywhere we use it and not only in the TTBR implementation. Thoughts?

src/NServiceBus.AmazonSQS/MessagePump.cs
```diff
@@ -421,14 +421,13 @@ static bool IsMessageExpired(Message receivedMessage, Dictionary<string, string>
             }
 
             var sentDateTime = receivedMessage.GetSentDateTime(clockOffset);
```
mikeminutillo: Is this already adjusting to local time? In the implementation for this method it is adding a clock offset to the `SentTimestamp` attribute. Is that attribute AWS Server time? If so then `sentDateTime` is already in local time and we don't need the corrected UTC, we need local UTC (which we already had).

danielmarbach: @mikeminutillo yes that attribute is the server time which gets adjusted with the client offset that is passed into the extension method.

danielmarbach:
There is also this UtcNow call https://github.com/Particular/NServiceBus.AmazonSQS/blob/master/src/NServiceBus.AmazonSQS/MessagePump.cs#L173

PR #308, opened 2019-06-10
Update Particular.Packaging to latest and add required LICENSE file

mauroservienti:


LICENSE.md
```diff
@@ -0,0 +1,10 @@
+By accessing NServiceBus code here, you are agreeing to the following licensing terms.
```
bording: This doesn't look like the right content. It should be the from RepoStandards.

mauroservienti: I’ll copy pasted from another repo, to get it from RepoStandards can I simply copy it from that repo?

bording: Yes, that's fine.

bording:
Looking at the build failure, there might some additional tweak needed to the GitVersion.yml to ignore a commit.

I'll take a look.

LICENSE.md
mauroservienti: Copied from RepoStandards (https://github.com/Particular/NServiceBus.AmazonSQS/pull/308/commits/121a14891d2f0e2c42f0765d8860007fec3abb36)

andreasohlund:
@mikeminutillo seems to have the same issue with the linux tests not reporting :(

PR #312, opened 2019-06-12
Normalize the way clock skew is handled when receiving messages

### LINKED ISSUE #210: TTBR implementation does not account for clock skew

mauroservienti:
replaces #307 and #311 
Fix #210


src/NServiceBus.AmazonSQS/MessagePump.cs
```diff
@@ -420,15 +423,15 @@ static bool IsMessageExpired(Message receivedMessage, Dictionary<string, string>
...
             var expiresAt = sentDateTime + timeToBeReceived;
-            if (expiresAt > utcNow)
+            if (expiresAt > DateTime.UtcNow)
             {
                 return false;
             }
+
             // Message has expired.
-            Logger.Info($"Discarding expired message with Id {messageId}, expired {utcNow - expiresAt} ago at {expiresAt} utc.");
+            Logger.Info($"Discarding expired message with Id {messageId}, expired {DateTime.UtcNow - expiresAt} ago at {expiresAt} utc.");
```
mikeminutillo: Should we be storing `Now` above and using it in both places?

mauroservienti: Good point @mikeminutillo, fixed.

PR #315, opened 2019-09-05
WIP Added support to ServerSideEncryptionMethod

imperugo:
Hi, 
we are using Sqs Transport combined with S3 for large messages. Our S3 storage is a private storage with encryption.
Looking the code I noticed that the implementation inside this package doesn't support that scenario, so I opened a PR.

My idea is to allow the user to specify the encryption method into the settings and use it into the `PutObjectRequest`.

Take a look and tell me if it does make sense to you (maybe you didn't implement it for a specific reason I don't know).

Thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,7 @@ installer/ServiceControl-cache
 *.userosscache
 *.sln.docstates
 .vs/
+.vscode/
 
 # mac temp file ignore
 .DS_Store
--- a/src/NServiceBus.AmazonSQS/Configure/S3Settings.cs
+++ b/src/NServiceBus.AmazonSQS/Configure/S3Settings.cs
@@ -11,7 +11,7 @@ namespace NServiceBus
     /// </summary>
     public class S3Settings : ExposeSettings
     {
-        internal S3Settings(SettingsHolder settings, string bucketForLargeMessages, string keyPrefix, string encryption) : base(settings)
+        internal S3Settings(SettingsHolder settings, string bucketForLargeMessages, string keyPrefix, string encryptionMethod) : base(settings)
         {
             Guard.AgainstNull(nameof(bucketForLargeMessages), bucketForLargeMessages);
             Guard.AgainstNullAndEmpty(nameof(keyPrefix), keyPrefix);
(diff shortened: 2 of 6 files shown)


danielmarbach:
Hi @imperugo 

Not an expert in this topic unfortunately but we recently fixed a bug that prevented users from using the transport with S3 and server side encryption, see https://github.com/Particular/NServiceBus.AmazonSQS/issues/291

When I tested it the code worked without having to specify the encryption setting like you did in this PR. Can you explain why you think this is needed?

danielmarbach:
I created a bucket with KMS

image

Then I run the simple SQS sample against that bucket

https://docs.particular.net/samples/sqs/simple/

and it worked fine. But again I'm not an expert in this area so I might be missing something

imperugo:
Hi @danielmarbach 
thanks for the answer. I've checked you document and the #291 but is another way to handle a similar problem.

As far as I know you could use the encryption into the storage using KMS or not. In case or not, my case, you have to send the encryption algorithm to the `IAmazonS3` request.
Moreover I can't change the encryption because it is used by tons of clients.

The idea behind this PR is that you can handle the encryption with KMS, as you mentioned, or in another way (not mandatory)

Exactly like you I'm not the biggest expert in this scenario because I can't manage my S3 storage configuration here, but I know that with this:

```
await s3Client.PutObjectAsync(new PutObjectRequest
                    {
                        BucketName = configuration.S3BucketForLargeMessages,
                        InputStream = bodyStream,
                        Key = key,
                        ServerSideEncryptionMethod = configuration.S3EncryptionMethod
                    }).ConfigureAwait(false);
```

it works (only adding the ServerSideEncryptionMethod property)

danielmarbach:
According to https://docs.aws.amazon.com/AmazonS3/latest/dev/kms-using-sdks.html it might also make sense to be able to configure the `ServerSideEncryptionKeyManagementServiceKeyId`

I'll raise an internal triage issue and we will come back to you as soon as possible. In the meantime thanks for the PR!

danielmarbach:
Raised https://github.com/Particular/NServiceBus.AmazonSQS/issues/316

danielmarbach:
Marked as WIP until we move forward with this

danielmarbach:
@imperugo I started looking into this and raised https://github.com/Particular/NServiceBus.AmazonSQS/pull/332 as an alternative. There is still some things to do there but what do you think about that proposal?

danielmarbach:
@imperugo replaced this PR with https://github.com/Particular/NServiceBus.AmazonSQS/pull/332 and released as 4.4.0. Can you please reach out to daniel dot marbach at particular dot net? I would like to discuss a small contribution gift with you.

Issue #318, opened 2019-10-02
Maximum message batch size calculation is incorrect, causing consistent failures when processing messages that result in multiple and/or large outgoing messages


rafalkozik:
### Symptoms

Processing of a message that results in sending out multiple large messages consistently fails part-way through. Some outgoing messages are dispatched to the broker but some are not. The message being processed is returned back to its queue and re-processed. Each processing attempt fails with the same problem (assuming processing logic is deterministic), eventually leading to moving that message to the error queue.

Retrying the message from ServiceControl does not solve the problem. The symptoms of the issue include:
- message not being processed on time
- duplicate messages being generated

### Who's affected

All users of SQS transport are potentially affected but the likelihood of the problem actually occurring depends on the number and size of messages being sent. 

### Root cause

The batch size calculation is wrong, leading NServiceBus to send a batch that exceeds the maximum size allowed by SQS. The batch size calculation does not take into account the message attributes.

### Original issue description

I have received exception after completion of the event handler on sending messages to SQS. There is S3 bucket configured for large messages. I've seen this issue only once out of 7 millions of messages processed by the endpoint.

NServiceBus.AmazonSQS: 4.3.4
AWSSDK.SQS: 3.3.3.10

`Amazon.SQS.Model.BatchRequestTooLongException: Batch requests cannot be longer than 262144 bytes. You have sent 262154 bytes. ---> Amazon.Runtime.Internal.HttpErrorResponseException: Exception of type 'Amazon.Runtime.Internal.HttpErrorResponseException' was thrown.
   at Amazon.Runtime.HttpWebRequestMessage.GetResponseAsync(CancellationToken cancellationToken)
   at Amazon.Runtime.Internal.HttpHandler``1.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.Unmarshaller.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.SQS.Internal.ValidationResponseHandler.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.ErrorHandler.InvokeAsync[T](IExecutionContext executionContext)
   --- End of inner exception stack trace ---
   at Amazon.Runtime.Internal.HttpErrorResponseExceptionHandler.HandleException(IExecutionContext executionContext, HttpErrorResponseException exception)
   at Amazon.Runtime.Internal.ErrorHandler.ProcessException(IExecutionContext executionContext, Exception exception)
[...]

PR #322, opened 2019-10-07
Attributes contribute to message size restrictions

danielmarbach:
Replaces https://github.com/Particular/NServiceBus.AmazonSQS/pull/319
Fixes https://github.com/Particular/NServiceBus.AmazonSQS/issues/318

### DIFF
--- a/src/NServiceBus.AmazonSQS/PreparedMessage.cs
+++ b/src/NServiceBus.AmazonSQS/PreparedMessage.cs
@@ -1,12 +1,13 @@
 namespace NServiceBus.Transports.SQS
 {
+    using System;
     using System.Collections.Generic;
     using Amazon.SQS.Model;
 
     class PreparedMessage
     {
         public string MessageId { get; set; }
-        public string Body{ get; set; }
+        public string Body { get; set; }
         public string Destination { get; set; }
         public string OriginalDestination { get; set; }
         public string QueueUrl { get; set; }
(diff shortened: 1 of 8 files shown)


danielmarbach:
This also need to extend into the batcher and properly cache the message size. So consider this as WIP

src/NServiceBus.AmazonSQS/PreparedMessage.cs
```diff
@@ -1,18 +1,64 @@
...
         public Dictionary<string, MessageAttributeValue> MessageAttributes { get; } = new Dictionary<string, MessageAttributeValue>();
         public string MessageGroupId { get; set; }
         public string MessageDeduplicationId { get; set; }
+
+        public long CalculateSize()
+        {
+            long size = Body.Length;
+            foreach (var messageAttributeValue in MessageAttributes)
+            {
+                size += messageAttributeValue.Key.Length;
```
rafalkozik: From my experiments it looks that _DataType_ counts as well, sending this: 

`	var sendMessageBatchRequest = new SendMessageBatchRequest();

	sendMessageBatchRequest.Entries.Add(new SendMessageBatchRequestEntry(Guid.NewGuid().ToString(), new string('x', 1024 * 256))
	{
		MessageAttributes = new Dictionary<string, MessageAttributeValue>
		{
			["foo"] = new MessageAttributeValue { DataType = "String", StringValue = "abcdef" },
			["bar"] = new MessageAttributeValue { DataType = "String", StringValue = "xyz" }
		},
	});`

Results in _One or more errors occurred. (Batch requests cannot be longer than 262144 bytes. You have sent 262171 bytes.)_, which is 27 bytes above the limit.

danielmarbach: Interesting. I reread the spec again and saw now

> The data type has the same restrictions as the message body (for more information)

It seems in my first read-through I thought it only applies for custom data types. Will fix

danielmarbach:
Replaced by https://github.com/Particular/NServiceBus.AmazonSQS/pull/329

Issue #46, opened 2015-12-07
Interest in using Amazon SNS to offer native broker capability?


chrissimon-au:
NServiceBus with Azure or RabbitMQ transport does not rely on persistent storage for subscription data, rather relying on the broker capability to manage subscriptions.

In AWS-land, SNS seems like the natural service to offer equivalent capability.  A rough outline of behaviour:
- When detecting event messages, check for existence of/create an SNS topic per event
- When subscribing, rather than sending a subscribe message to the 'owner' endpoint, subscribe the subscriber's input-queue to the SNS topic.

Is this something that you guys have already been thinking about?

ahofman:
Thanks @chrissimon-au for your great suggestion! This isn't something I've considered yet, but your rough outline appears to be a great fit. 

I unfortunately don't get a great deal of time to work on this plugin at the moment - feel free to create a PR and have a go at building it yourself if you want. If you decide to do so, please assign yourself to this issue (and if I ever do get the time to implement this, I'll do the same). 

(please forgive the lateness of my reply).

chrissimon-au:
Hi @ahofman - no problems, I know how it gets!

We are working on a project right now where this might be very useful - will let you know if I get a chance to work on it!

ahofman:
A few questions about your project to help me prioritise this work:

1. What kind of timelines are you looking at for your project? i.e., when would you need this feature implemented?
2. If this feature wasn't implemented, would it be a killer for your project? i.e., would you have to use a different transport?

chrissimon-au:
Hi Adrian,

Thanks for following up!  Our project is in play this year - we are likely to have to launch (MVP) in the near future (next few months) - the absence of this certainly wouldn't be a killer for the project - we'd stick with SQS transport and traditional persistence based subscriptions.

If we get a chance to implement SNS subscriptions (or you do) we'd likely switch to that after launch though, as I see a number of advantages into the future (including non NServiceBus subscribers, and auditing all published events even if no subscribers...)

Cheers,
Chris.

biddlem:
Is there any traction on this feature yet?  This would be extremely helpful for the exact reasons that @chrissimon-au  mentioned.

mauroservienti:
Hi @biddlem. We're working on providing support for "_native_" pub/sub for the SQS transport without the need for SNS. The goal is to allow users to use the transport without the need for external subscription storage and, at the same time, without the need to use another service, which means more costs, more moving parts, and more configuration.

[...]

PR #430, opened 2020-03-30
Custom mapping API proposal

mauroservienti:
While working on the SNS implementation we realized that the original idea of having topics created for each type in the message hierarchy is problematic. For this reason, we're proposing a change in the way inheritance works when subscribing to events and make inheritance support opt-in and not enabled by default as previously planned.

By default subscribers will create and subscribe to topics only for the most concrete type they know about. Users will have configuration options to manually influence this behavior.

Relevant changes are in the following files:

- [src/NServiceBus.AmazonSQS/Configure/SqsTransportSettings.cs](https://github.com/Particular/NServiceBus.AmazonSQS/pull/430/files#diff-cea21dc686157fa9d0499762093f3dd9)
- [src/NServiceBus.AmazonSQS/Configure/EventToTopicsMappings.cs](https://github.com/Particular/NServiceBus.AmazonSQS/pull/430/files#diff-5e459a429ccb70a3979f5376ae4271b6)
- [src/NServiceBus.AmazonSQS/Configure/SubscriptionManager.cs](https://github.com/Particular/NServiceBus.AmazonSQS/pull/430/files#diff-4ca408a0ce568eaa91e76f60fa9fae99)
- [src/NServiceBus.AmazonSQS/TransportConfiguration.cs](https://github.com/Particular/NServiceBus.AmazonSQS/pull/430/files#diff-864011499c67711bda4277ba76052ead)

## Sample scenario

Using the following as a sample message structure:

```
interface ISalesEvent{ }

class OrderAccepted : ISalesEvent{}
```

And a subscriber with a handler like:

```
class OrderAcceptedHandler : IHandleMessages<OrderAccepted>{}
```

A topic named `namespace-OrderAccepted` will be created, by the subscriber, and a subscription will be set.

NOTE: topics names can be fully customized and the above default behavior that uses the full type name is still under evaluation and may be subject to change before release.

If the subscriber, however, wants to subscriber to `ISalesEvent` and the publisher still publishes a concrete `OrderAccepted` message, the automatic subscription behavior doesn't work and manual configuration is required at the subscriber side:

```
var endpointConfig = new EndpointConfiguration("subscriber-endpoint");
var transportConfig = endpointConfig.UseTransport<AmazonSQS>();
transportConfig.MapEvent<ISalesEvent>(topic: "namespace-OrderAccepted");
```

The above custom topic mapping allows the following handler to work as expected:

```
class SalesEventHandler : IHandleMessages<ISalesEvent>{}
```

## Issues with the original plan

The original plan was to implement a `topic-per-message-type` topology where for each message type, and for each type in the inheritance chain, a topic would have been created.

NOTE: SNS topics cannot be subscribed to other topics as, for example, it can be done in RabbitMQ

For example, if message classes are defined as in the above sample the following topics would have been created:

- `namespace-ISalesEvent`
- `namespace-OrderAccepted`

If a subscriber defines a message handler like the following:

```
class OrderAcceptedHandler : IHandleMessages<OrderAccepted>{}
```

during startup, it would create a subscription to the `namespace-OrderAccepted` topic. However, if the same subscriber defines also a handler like:

```
class SalesEventHandler : IHandleMessages<ISalesEvent>{}
```

The subscriber cannot subscribe to the `namespace-ISalesEvent` topic too.

The reason for this lies in the publisher. When publishing, a publisher has no way to know who is subscribed to what, so when publishing a message like `OrderAccepted` the only option is to publish it to all known topics in the message inheritance chain. The publishing process, in this case, would need to:

- inspect the message inheritance chain
- for each type in the chain compute the topic name
- publish to each topic

This behavior has a few concerning side effects:

- Given that there are no transactions in SQS/SNS the publisher might fail to publish to some of the topics leading to a corrupted/message loss scenario
- subscribers containing handlers for more than one type in the inheritance chain will receive duplicates by definition making handling code way much more complex

Even if it's true that the last issue could be solved by the NServiceBus.Outbox we feel that requiring the outbox even for simple pub/sub scenarios is not ideal. The first problem has no solution and way much worse consequences.

One solution is to use SNS filtering policies on a single topic to "route" topics messages to subscribers. The issue is that, even if filtering policies are quite powerful, they are limited in the amount available. By default, there can be only 200 filtering policies, and the limit is per account per region. Even if 200 was enough, we cannot assume we own the whole available set of filtering policies.

Another solution to these issues could be to have subscribers subscribe only to the most abstract type they know about. Using the above sample the subscriber would then subscribe to `namespace-ISalesEvent` topic only, and the NServiceBus support for polymorphic dispatch will take care of dispatching the message the all the relevant handlers. Even if this option apparently solves the duplication problem, publishers will still be required to publish to multiple topics.

The problem is that this second option introduces a much larger pain point when the system needs to evolve and the handlers topology changes. At a certain point in the system evolution, for example, the following handler could be removed:

```
class SalesEventHandler : IHandleMessages<ISalesEvent>{}
```

At startup now the subscriber needs to compute how to change the subscriptions to match the new handlers. Even if it was possible to compute all the possible changes permutations, changes cannot be then applied because other instances of the same subscriber might not have been updated yet. This would make deploying endpoints with no downtime impossible.


src/NServiceBus.AmazonSQS/Configure/EventToTopicsMappings.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            if (!eventsToTopicsMappings.TryGetValue(eventType, out var mapping))
+            {
+                mapping = new HashSet<string>();
+                eventsToTopicsMappings.Add(eventType, mapping);
+            }
+
+            foreach (var topicName in topicsNames)
+            {
+                if (!mapping.Contains(topicName))
```
danielmarbach: Pretty sure the hashset does that for us and makes the contains call redundant

src/NServiceBus.AmazonSQS/Configure/EventToTopicsMappings.cs
```diff
@@ -0,0 +1,38 @@
...
+    using System.Collections.Generic;
+    using System.Linq;
+
+    class EventToTopicsMappings
+    {
+        public void Add(Type eventType, string[] topicsNames)
+        {
+            if (!eventsToTopicsMappings.TryGetValue(eventType, out var mapping))
+            {
+                mapping = new HashSet<string>();
```
danielmarbach: If you specify the comparer in the ctor the below contains code is no longer necessary I think

src/NServiceBus.AmazonSQS/Configure/EventToTopicsMappings.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        public bool HasMappingsFor(Type eventType)
+        {
+            return eventsToTopicsMappings.ContainsKey(eventType);
+        }
+
+        public string[] GetMappedTopicsNames(Type eventType)
+        {
+            return eventsToTopicsMappings[eventType].ToArray();
```
danielmarbach: I would just return  the HashSet or one of the interfaces implemented by the hashset instead of penalizing ourself with an array alloc everytime we call this method

mauroservienti:
> It would be great to see in this PR how the existing failing acceptance tests leverage the new API to make them pass

yeah. It's coming, I wanted to get the API design first.

timbussmann:
> Another solution to these issues could be to have subscribers subscribe only to the most abstract type they know about. Using the above sample the subscriber would then subscribe to namespace-ISalesEvent topic only, and the NServiceBus support for polymorphic dispatch will take care of dispatching the message the all the relevant handlers. Even if this option apparently solves the duplication problem, publishers will still be required to publish to multiple topics.

Have you considered to enforce only a single hierarchy chain (and throw if violating this constraint)?
Assuming the publisher would also only publish to the most abstract (public) type, in what case would the publisher have to publish to multiple topics? In that case, the sample you're describing shouldn't cause problems?

On your suggested API:
> transportConfig.MapEvent<ISalesEvent>(topic: "namespace-OrderAccepted");

have you considered to make this use the target type instead to make it easier for the user and not expect him to know what queue name a certain type is converted to, e.g. ` transportConfig.MapEvent<ISalesEvent, OrderAccepted>();`

danielmarbach:
```
transportConfig.MapEvent<ISalesEvent, OrderAccepted>();
```

we have not yet. This is exactly why we wanted to open for input before we go down the path of too much baking things into stone. So this input is greatly appreciated

>Have you considered to enforce only a single hierarchy chain (and throw if violating this constraint)?

We have evidence from interactions like https://github.com/Particular/NServiceBus.AmazonSQS/issues/46#issuecomment-584447851 as well as "it was supported before we should allow transition kind of thoughts" that it is good to support it. That's why we opted for an explicit mapping API. That gives us also the ability to deprecate the mapping API in the future should we decide to move away from inheritance as a whole from the platform.

> Assuming the publisher would also only publish to the most abstract (public) type, in what case would the publisher have to publish to multiple topics? In that case, the sample you're describing shouldn't cause problems?

We have to chew on this before we can give you an answer. Only while implementing and revaluation our plans we found out the most concrete type strategy. The text you referred to was written in the original plan under the assumptions of the plan before.

mauroservienti:
> Have you considered to enforce only a single hierarchy chain (and throw if violating this constraint)?

We discussed it and in the original plan we found no automatic way to support multiple inheritance. Making what you’re suggesting the default behavior. Single inheritance has always been in the plans with the described caveats, that’s the reason for this proposal.
 > Assuming the publisher would also only publish to the most abstract (public) type, in what case would the publisher have to publish to multiple topics? In that case, the sample you're describing shouldn't cause problems?

Let’s assume publisher publishes `OrderAccepeted` to one `namespace-OrderAccepted ` topic only. If the subscriber is subscribed to `ISalesEvent` only it’ll have no way to discover to which topic it should subscribe to, because the subscriber has no knowledge of the existence of the `OrderAccepted` derived type.

Or am I missing something Tim?

danielmarbach:
In addition to that consumer driven contracts add additional headache to the picture which brings us full speed back to the mapping API

jmenziessmith:
Thanks for doing this! One reason we are keen to use SNS is to enable interoperability between other systems that may not use NServiceBus. In this scenario, topics per interface are a little less relevant and we are really interested in the concrete types (domain events).

Admittedly our use case may not cover all scenarios, but one topic per type would work for us. Should there be multiple types that all implement the same interface, then surely you could use reflection to subscribe all respective topics based on the set of concrete types implementing such interface?

Would that avoid the need to manually add mappings?

We have actually ended up building an interim solution that uses middleware to publish to sns, wrapping the message body in a non NServicebus envelope. To consume the messages we implemented the subscription mechanism to get sns messages onto the nsb queue and added  a message mutator when consuming inbound messages to convert back from the generic envelope to an nsb format message.

timbussmann:
> Let’s assume publisher publishes OrderAccepeted to one namespace-OrderAccepted topic only. If the subscriber is subscribed to ISalesEvent only it’ll have no way to discover to which topic it should subscribe to, because the subscriber has no knowledge of the existence of the OrderAccepted derived type.

I'd assume the publisher would have to publish the event to the most-abstract type queue as well, therefore it would publish to the ISalesEvent-queue. If both subscriber and publisher use the same type, they should both be able to find the same most-abstract type? It would add some complexity to the client to be able to identify messages implementing the same abstract type but without having a handler for those and discarding them though. I'm not proposing this as an alternative, I was just pointing out that the described problems in your proposal might not be an issue under certain circumstances.

mauroservienti:
> > Let’s assume publisher publishes OrderAccepeted to one namespace-OrderAccepted topic only. If the subscriber is subscribed to ISalesEvent only it’ll have no way to discover to which topic it should subscribe to, because the subscriber has no knowledge of the existence of the OrderAccepted derived type.
> 
> I'd assume the publisher would have to publish the event to the most-abstract type queue as well, therefore it would publish to the ISalesEvent-queue. If both subscriber and publisher use the same type, they should both be able to find the same most-abstract type? It would add some complexity to the client to be able to identify messages implementing the same abstract type but without having a handler for those and discarding them though. I'm not proposing this as an alternative, I was just pointing out that the described problems in your proposal might not be an issue under certain circumstances.

@timbussmann we considered this scenario during the brainstorming phase, it was discarded as it has the risk of putting to much pressure on subscribers due to the need of consuming messages from the input queue only to later discard them. It might also have a significant impact on costs. This scenario would have been ideal if filtering polices were a solution.

mauroservienti:
@jmenziessmith 

> Thanks for doing this! One reason we are keen to use SNS is to enable interoperability between other systems that may not use NServiceBus. In this scenario, topics per interface are a little less relevant and we are really interested in the concrete types (domain events).

Integration is one of the reasons, if not the main one, that drove us in the decision to implement SNS support.

> 
> Admittedly our use case may not cover all scenarios, but one topic per type would work for us. Should there be multiple types that all implement the same interface, then surely you could use reflection to subscribe all respective topics based on the set of concrete types implementing such interface?
> 
> Would that avoid the need to manually add mappings?

This is tricky as in .NET it's easy to walk the inheritance chain up but not down.
What I mean by this is that we may end up in a situation where the following messages are defined:

- `ISalesEvent` in the Contracts assembly
- `OrderAccepted` in the SalesMessages assembly

If the subscriber has a handler defined as:

```
class SalesEventHandler : IHandleMessages<ISalesEvent>{}
```

For it to work a reference to both of the assemblies is needed and the subscriber should scan all the types available in all assemblies in the bin folder and look for all the message types that implement the `ISalesEvent` interface, once all the "most" concrete types are found the subscriber can determine which topics it needs to subscribe to.

This could work, and it's an interesting approach we should explore a bit further.

The problem I see out of the top of my head is: how do we handle diffs when the handlers topology changes? This is especially important in a scaled-out environment where there are multiple instances of the same endpoint that could be upgraded not all at the same time.

Thanks for the hint, it's worth exploring further.

src/NServiceBus.AmazonSQS/Configure/EventToTopicsMappings.cs
mauroservienti: Not sure I understand @danielmarbach

PR #437, opened 2020-04-09
Roll once

danielmarbach:


mauroservienti:
@danielmarbach does this need a way to discard leftovers from previous tests run? e.g. a test session is run, for reasons messages are not consumed from destination endpoints input queues, a subsequent test run will receive messages from the previous run.

Should there be something like an `aws-test-session-header` appended to outgoing messages and used by receivers to discard incoming messages not pertinent to the test session?
Or, does `PurgeMessagesOnStartup(true)` would do the trick?

danielmarbach:
Closed in favor of https://github.com/Particular/NServiceBus.AmazonSQS/pull/441

PR #518, opened 2020-06-29
Reschedule and deliver timeouts using batching

### LINKED ISSUE #519: Transport fails to reschedule and deliver unrestricted delayed messages under heavy load
## Who's affected

Customers using unrestricted delayed delivery with many timeouts expiring at the same time

## Symptoms

If there is a large number of delayed messages the AmazonSQS transport might fail in dispatching or in rescheduling them due to requests throttling by the SQS broker. The underlying issue is caused by the unrestricted native delayed deliveries behavior in the message pump. When a message is delayed the following happens:

- the message is sent to the local `-delay.fifo` queue with a custom header that specifies the delay length
- if the delay length is shorter than or equal to 15 minutes (900 seconds) the message is then moved to the main input queue
- if the delay is longer the message is rescheduled again till the timeout expires and the message is finally moved to main input queue for consumption

If there is a large number of delayed messages like in the [mentioned discussion](https://discuss.particular.net/t/request-is-throttled-for-delayed-messages/1742) the message pump can get throttled and will start failing in rescheduling timeouts and in delivering expired timeouts.

Limitations when using a FIFO queue are spelled out in [AWS documentation](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html) and when using FIFO queues there is a limit of 300 messages per second. That limit can be overcome by batching requests to the `-delay.fifo` queue up to 300 batch requests per second. Given that each batch can contain up to 10 messages this means that by batching those requests the transport could be able to reschedule up to 3,000 timeouts per second.

## Backported to

- 5.0.1
- 4.4.1

danielmarbach:
Fixes #519

This introduced batched operations for natively delayed messages. This brings down significantly the operation costs and reduces the likelihood of getting throttled when a larger number of delayed operations are due.

> By default, with batching, FIFO queues support up to 3,000 messages per second (TPS), per API action (SendMessageBatch, ReceiveMessage, or DeleteMessageBatch).

> Q: Do Amazon SQS batch operations cost more than other requests?
> No. Batch operations (SendMessageBatch, DeleteMessageBatch, and ChangeMessageVisibilityBatch) all cost the same as other Amazon SQS requests. By grouping messages into batches, you can reduce your Amazon SQS costs.

https://aws.amazon.com/sqs/faqs/

## Design decisions

- Even though some code closely resembles the code for the message dispatcher the only thing that is shared across the two paths is the batcher logic. The actual dispatching logic is not shared because we expect those two cases to be distinct enough so that sharing code could hinder our ability to evolve the two paths invidually
- The pump has been split into two parts
  - The input queue pump contains the logic for handling messages on the input queue
  - The delayed messages pump that takes care of fetching delayed messages and rescheduling or sending them to the input queue
- Purging is now done earlier, we think this should have no operational impact


danielmarbach:
LGTM

PR #544, opened 2020-08-11
Resolve failing test

mikeminutillo:
This test attempts to subscribe to two different events and the transport under the covers can take quite a long time to do that because of https://github.com/aws/aws-sdk-net/issues/1569

Unfortunately this results in the test failing because it exceeds the timeout given to test setup outlined here https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs#L144-L146

The test doesn't need to subscribe after startup on this transport. It only needs that if the transport relies on message-driven pub-sub (to ensure that we record when the subscribers are all wired up). For the current version of SQS, the autosubscribe feature is able to set things up for us at endpoint startup, and we take advantage of the fact that we know we're single threaded when we do that.


mikeminutillo:
I copied the test out of core and into the repo to modify it to rely on autosubscribe

PR #551, opened 2020-08-24
Obsolete with warning the message-driven pubsub compatibility mode

danielmarbach:


mauroservienti (approved):
Does this documentation need to be tweaked https://docs.particular.net/transports/sqs/configuration-options#custom-topics-mappings-message-driven-pubsub-compatibility-mode?

danielmarbach:
My thinking was only after we obsolete with error because right now they can still use it or do you see reasons why we should say somethng like

As of 5.1 the API is obsoleted...?

PR #552, opened 2020-08-25
Avoid casting down the transport infrastructure and use reflection to access the factory field instead

SzymonPobiega:
Version 5 of the transport introduced a new class `SettlePolicy` for the purpose of executing all the autosubscribe requests in a single go. This has been added because executing them concurrently without batching caused resulted in overwriting the policy.

The `SettlePolicy` obtained reference to the transport infrastructure object and attempted to downcast it to the specific SQS transport infrastructure type. This broke the serverless packages which rely on the ability to wrap the transport infrastructure in order to substitute the message pump with a no-op implementation. 

This change replaces the downcast with usage of public API `ConfigureSubscriptionInfrastructure()`, taking advantage of the fact that the implementation of the subscription manager factory method always returns the same instance. Unfortunately the `SubscriptionManagerFactory` property is internal so it needs to be invoked via reflection. BTW, the same hack is used by NServiceBus.Raw.

### DIFF
--- a/src/NServiceBus.Transport.SQS/Configure/SqsTransportInfrastructure.cs
+++ b/src/NServiceBus.Transport.SQS/Configure/SqsTransportInfrastructure.cs
@@ -126,12 +126,6 @@
 
         public override TransportSubscriptionInfrastructure ConfigureSubscriptionInfrastructure()
         {
-            //NServiceBus core only calls this method if endpoint is not send-only. The null path is only used by SettlePolicy class
-            //that retrieves the subscription manager in order to call Settle on it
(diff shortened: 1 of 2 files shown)


danielmarbach (approved):
The only thing I would add is some more background info about the things we considered and did not pursue to the description of the PR

danielmarbach:
Discussed with @SzymonPobiega and we are going to align the settle feature with the autosubscribe precondition instead of doing the check in the transport infrastructure

https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs#L20

PR #576, opened 2020-10-19
Update to 3.5 SDK and adjust policy json conversion

danielmarbach:


danielmarbach:
Unfortunately, the Policy structure adapts to whether there is a single statement or multiple statements present and thus the current structure in this PR doesn't reflect that.

I raised https://github.com/Particular/NServiceBus.AmazonSQS/pull/577

PR #577, opened 2020-10-19
Update SDK to 3.5 and suppress policy usage

danielmarbach:
Alternative to https://github.com/Particular/NServiceBus.AmazonSQS/pull/576 which unblocks 3.5 SDK

The policy objects are still there and still fully functional (they are still in use in SQS as well). So until they are really removed I guess the best way is to pragma disable the warning. Once the SDK itself has been adjusted we hopefully have a good way we can borrow from the SDK to settle the policy again or maybe https://github.com/aws/aws-sdk-net/issues/1569 has been resolved


mauroservienti (approved):
Do we an idea about what will replace policies?

danielmarbach:
Not really, the documentation is extremely vague, see

https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies.html#access_policies-json

I've raised an issue https://github.com/aws/aws-sdk-net/issues/1732

PR #600, opened 2020-12-18
Enable native messages to be processed and if needed sent through recoverability

lailabougria:
Requires the following:
- Required: A message attribute "MessageTypeFullName" containing the full name of the message type
- Optional: When the payload is on S3, a message attribute with key "S3BodyKey" so that we can download it
- if the message body is not on S3, it should be base 64 encoded and serialized using the technique the endpoint is configured (default XML)


danielmarbach (approved):

src/NServiceBus.Transport.SQS.AcceptanceTests/NativeIntegration/When_receiving_a_native_message.cs
```diff
@@ -0,0 +1,203 @@
...
+                    {
+                        c.CustomConfig((cfg, ctx) =>
+                        {
+                            ctx.ErrorQueueAddress = cfg.GetSettings().ErrorQueueAddress();
+                        });
+                        c.When(async (session, ctx) =>
+                        {
+                            await SendNativeMessage(new Dictionary<string, MessageAttributeValue>
+                            {
+                                // unfortunately only the message id attribute is preserved when moving to the poison queue
```
danielmarbach, line 59: Move to error queue does the following

```
                await sqsClient.SendMessageAsync(new SendMessageRequest
                {
                    QueueUrl = errorQueueUrl,
                    MessageBody = message.Body,
                    MessageAttributes =
                    {
                        [Headers.MessageId] = new MessageAttributeValue
                        {
                            StringValue = messageId,
                            DataType = "String"
                        }
                    }
                }, CancellationToken.None).ConfigureAwait(false);
                // The MessageAttributes on message are read-only attributes provided by SQS
                // and can't be re-sent. Unfortunately all the SQS metadata
                // such as SentTimestamp is reset with this send.
```

I have a hunch the comment is not correct and we can actually resent the attributes easily by doing this

```
                MessageAttributes = message.MessageAttributes.ToDictionary(x => x.Key, x => new SnsMessageAttributeValue
                {
                    DataType = x.Value.DataType,
                    StringValue = x.Value.StringValue,
                    BinaryValue = x.Value.BinaryValue,
                })
```

because right now moving to poison queue looses all attributes except the message id. Let's discuss

danielmarbach (approved):

src/NServiceBus.Transport.SQS/DelayedMessagesPump.cs
```diff
@@ -168,14 +169,17 @@ IReadOnlyCollection<SqsReceivedDelayedMessage> PrepareMessages(CancellationToken
...
+                            [DelaySeconds] = new MessageAttributeValue
                             {
                                 StringValue = remainingDelay.ToString(),
                                 DataType = "String"
                             }
                         }
                     };
 
+                    // for native integration scenario's we copy over all the message attributes so we don't lose part of the message when moving to the delayed delivery queue
+                    preparedMessage.CopyMessageAttributes(receivedMessage.MessageAttributes);
```
yvesgoeleven, line 181: This seems to be done unconditionally, is it safe to do on regular messages?

danielmarbach: I do think we always have to do it and we should probably slightly tweak the text. Because in any case we want to make sure when we fetch `*` we also want to copy all the attributes every that are necessary. Thoughts @lailabougria

src/NServiceBus.Transport.SQS/DelayedMessagesPump.cs
```diff
@@ -168,14 +169,17 @@ IReadOnlyCollection<SqsReceivedDelayedMessage> PrepareMessages(CancellationToken
...
-                            [TransportHeaders.DelaySeconds] = new MessageAttributeValue
+                            [DelaySeconds] = new MessageAttributeValue
                             {
                                 StringValue = remainingDelay.ToString(),
                                 DataType = "String"
                             }
                         }
                     };
 
+                    // for native integration scenario's we copy over all the message attributes so we don't lose part of the message when moving to the delayed delivery queue
```
danielmarbach, line 180:
```suggestion
                    // We copy over all the message attributes so we don't lose part of the message when moving to the delayed delivery queue
```

src/NServiceBus.Transport.SQS/DelayedMessagesPump.cs
```diff
@@ -194,6 +198,10 @@ IReadOnlyCollection<SqsReceivedDelayedMessage> PrepareMessages(CancellationToken
                         QueueUrl = inputQueueUrl
                     };
 
+                    // for native integration scenario's we copy over all the message attributes so we don't lose part of the message when moving to the delayed delivery queue
```
danielmarbach, line 201:
```suggestion
                    // We copy over all the message attributes so we don't lose part of the message when moving to the delayed delivery queue
```

danielmarbach:
@yvesgoeleven added an edit suggestion

src/NServiceBus.Transport.SQS/DelayedMessagesPump.cs
lailabougria: Yes exactly, especially due to the fact that we create a mix of attributes on the native message attributes and some in the header dictionary once a message goes through retries. The first time we'd be handling a native message, the second time, a transportmessage.

yvesgoeleven (approved):

PR #604, opened 2021-01-12
Batch the policy settlement even for hierarchies

### LINKED ISSUE #605: Subscribing to mapped event types might not set the policy correctly causing the published message to not be delivered to the subscriber
## Who's affected

- All customers using version 5.x of the transport, manually subscribing (`session.Subscribe<MyEvent>()`) to events that are mapped using [event topic mappings](https://docs.particular.net/transports/sqs/configuration-options#custom-topics-mappings)

## Symptoms

When manually subscribing with for example `await session.Subscribe<MyEvent>()` in combination with an [event topic mapping](https://docs.particular.net/transports/sqs/configuration-options#custom-topics-mappings) like

```
var transport = endpointConfiguration.UseTransport<SqsTransport>();
transport.MapEvent<MyEvent, PublishedEvent>();
```

the policy for the endpoint input queue might not be set correctly causing the published message not to be delivered to the subscriber.

danielmarbach:
Fixes #605

image

With these changes indeed the test has never failed again so far.

### DIFF
--- a/src/NServiceBus.Transport.SQS.Tests/SubscriptionManagerTests.cs
+++ b/src/NServiceBus.Transport.SQS.Tests/SubscriptionManagerTests.cs
@@ -181,6 +181,8 @@ namespace NServiceBus.Transport.SQS.Tests
         [Test]
         public async Task Subscribe_with_event_to_events_mapping_settles_policy_once_instead_of_for_all_topics()
         {
+            var manager = CreateNonBatchingSubscriptionManager();
+
             var subscribedEventType = typeof(IEvent);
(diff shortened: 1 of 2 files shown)


danielmarbach:
@mauroservienti I only added you as a reviewer because we have recently been talking about this. Feel free to ignore when you have more pressing things on your plate

mauroservienti (approved):

mauroservienti:
Clean and straightforward implementation. Good job.

danielmarbach:
Rebased on master

PR #601, opened 2021-01-05
Switch policy settlement to use one statement with multiple conditions

danielmarbach:
This PR introduces a transition to how we create policies. 

## Explicit policy with all topics that are subscribed to

Instead of writing a policy statement per event that is subscribed to

```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "180523c7ef74469fbe8b166b8178c2cc",
      "Effect": "Allow",
      "Principal": {
        "AWS": "*"
      },
      "Action": "sqs:SendMessage",
      "Resource": "arn:aws:sqs:eu-central-1:714232877752:subscriber2",
      "Condition": {
        "ArnLike": {
          "aws:SourceArn": "arn:aws:sns:eu-central-1:714232877752:Messages-MyEvent"
        }
      }
    },
    {
      "Sid": "Sid1594040661602",
      "Effect": "Allow",
      "Principal": {
        "AWS": "*"
      },
      "Action": "SQS:SendMessage",
      "Resource": "arn:aws:sqs:eu-central-1:714232877752:subscriber2",
      "Condition": {
        "ArnLike": {
          "aws:SourceArn": "arn:aws:sns:eu-central-1:714232877752:Messages-MyOtherEvent"
        }
      }
    }
  ]
}
```

the policy gets condensed down to

```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "4eeb64b1326f4b658dd8b54a3b183491",
      "Effect": "Allow",
      "Principal": {
        "AWS": "*"
      },
      "Action": "sqs:SendMessage",
      "Resource": "arn:aws:sqs:us-east-1:714232877752:AT01SHG3XEJ0EIIYBIRIANUWBaseEventFrom2Publishers-Subscriber1",
      "Condition": {
        "ArnLike": {
          "aws:SourceArn": [
              "arn:aws:sns:eu-central-1:714232877752:Messages-MyEvent",
              "arn:aws:sns:eu-central-1:714232877752:Messages-MyOtherEvent"
          ]
        }
      }
    }
  ]
}
```

Additionally, it is possible to switch to wildcard policies by using the advanced policy API

```
var policy = transport.Policies();
policy.AddAccountCondition();
policy.AddTopicNamePrefixCondition(); // extracted from transport.TopicNamePrefix("DEV-")
policy.AddNamespaceCondition("Sales."); // dots turned to dashes and if prefix set it would be taken into account
[...]

PR #612, opened 2021-01-14
CLI Tool: Allow setting explicit event or wildcard based policies

yvesgoeleven:
CLI changes related to https://github.com/Particular/NServiceBus.AmazonSQS/pull/601

Introduces 3 new commands

> sqs-transport endpoint set-policy  "endpointname" events --event-type "event-type1" --event-type "event-type2" [--other-options] 

> sqs-transport endpoint set-policy "endpointname" wildcard --account-wildcard --namespace "namespacename" --prefix "prefix" --remove-event-type "event-type1"  --remove-event-type "event-type2"  [--other-options] 

> sqs-transport endpoint list-policy  "endpointname" [--other-options]

Changed the behavior of subscribe to no longer auto create policy

> sqs-transport endpoint subscribe name event-type [--other-options]

### DIFF
--- a/src/CommandLine/Endpoint.cs
+++ b/src/CommandLine/Endpoint.cs
@@ -94,9 +94,9 @@
             }
         }
 
-        public static async Task SetEventsBasedPolicy(IAmazonSQS sqs, IAmazonSimpleNotificationService sns, string prefix, string endpointName, IEnumerable<string> eventTypes)
+        public static async Task SetPolicy(IAmazonSQS sqs, IAmazonSimpleNotificationService sns, string prefix, string endpointName, IEnumerable<string> eventTypes, bool addAccountCondition, bool addPrefixcondition, IReadOnlyList<string> namespaceConditions)
         {
-            await Console.Out.WriteLineAsync($"Setting event based delivery policies on endpoint '{endpointName}'.");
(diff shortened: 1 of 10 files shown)


danielmarbach:
@yvesgoeleven Pushed a few minor cleanups and rebased on top of master

On thing that I find slightly "noisy" is all the in-between asserts but given that this is my personal preference only and I see that it allows you to retest certain things while not needing to explode the high-level test cases I'll approve

src/CommandLine/Endpoint.cs
```diff
@@ -92,5 +99,63 @@ public static async Task Unsubscribe(IAmazonSQS sqs, IAmazonSimpleNotificationSe
...
+            var policy = queueAttributes.ExtractPolicy();
+
+            if (!policy.Update(policyStatements,
+                addAccountCondition, 
+                addPrefixcondition,
+                namespaceConditions,
+                prefix,
+                queueArn))
+            {
+                await Console.Out.WriteLineAsync($"Policy on endpoint '{endpointName}' not set.");
```
danielmarbach, line 133: Maybe we could tone this down a bit to not give the impression something is off. How about something like

```suggestion
                await Console.Out.WriteLineAsync($"No updates needed for policy on endpoint '{endpointName}'.");
```

danielmarbach (approved):

lailabougria (approved):

PR #674, opened 2021-03-05
Support CancellationToken without onComplete - core alpha 881

DavidBoike:


DavidBoike:
The failing test is inexplicable and only happens on TeamCity. Hoping it just goes away when I re-run on Monday.

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -66,14 +66,16 @@ public async Task Initialize(PushRuntimeSettings limitations, OnMessage onMessag
             this.onError = onError;
         }
 
-        public Task StartReceive(CancellationToken cancellationToken)
+        public Task StartReceive(CancellationToken cancellationToken = default)
         {
-            if (tokenSource != null)
+            if (messagePumpCancellationTokenSource != null)
```
andreasohlund, line 71: This is the only of our pumps that guard for multiple calls to start, should we add this to all the others?

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -100,20 +102,21 @@ public Task StartReceive(CancellationToken cancellationToken)
...
             }
 
             return Task.CompletedTask;
         }
 
-        public async Task StopReceive(CancellationToken cancellationToken)
+        public async Task StopReceive(CancellationToken cancellationToken = default)
         {
-            if (tokenSource == null)
+            if (messagePumpCancellationTokenSource == null)
```
andreasohlund, line 113: This is the only transport that guards for multiple calls to stop, should we align the others?

andreasohlund (approved):

src/NServiceBus.Transport.SQS/InputQueuePump.cs
DavidBoike: Core already makes stopping a stopped endpoint a no-op: https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs#L25-L28

If we're going to spend cycles aligning anything, I'd like it to be in removing the duplicate code in the first place.

src/NServiceBus.Transport.SQS/InputQueuePump.cs
DavidBoike: Same thing as for Stop: https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Hosting/HostCreator.cs#L75-L82

PR #718, opened 2021-03-26
Bump SDK refs

danielmarbach:


DavidBoike (approved):

adamralph:
@danielmarbach @DavidBoike should this labelled so it lands in release notes?

danielmarbach:
When I'm releasing a component from master I do anyway always a compare to the latest release to see if something needs to be labeled, added to the milestone or whatever because that is safer than relying on people doing it when the PRs are merged in almost all the cases. So from my perspective it is not necessary

PR #751, opened 2021-04-12
Use dependency ranges con CLI

mauroservienti:


mauroservienti:
@danielmarbach do you think the same should apply to

- `McMaster.Extensions.CommandLineUtils`
- `Microsoft.Bcl.AsyncInterfaces`

Or maybe was there a reason for the CLI to not use ranges?

mauroservienti:
Let me rephrase the question because the reason is clear. There is no point in the CLI in using ranges otherwise it'll always be built against the lowest ones. However not using ranges means that dependabot will update the CLI project on `master`. Are we okay with that?

danielmarbach:
In general we tried to avoid ranges in the command line tool because of the reason you mentioned and also to avoid surprises because technically every installation of the tool might resolve a different version. For the tool we mostly care about having a specific version we know that does the job. That's it. The dependency does not affect the user of the command line because they do not depend upon the command line directly. Like many other command lines the dependencies get updated once the tool is released together with the other components as part of the same repository.

mauroservienti:
Thanks @danielmarbach. Closing.

PR #787, opened 2021-04-27
GithubActions - release-5.3

danielmarbach:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,58 @@
...
+      - release-[0-4].*
+  workflow_dispatch:
+env:
+  DOTNET_NOLOGO: true
+  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+  AWS_REGION: ${{ secrets.AWS_REGION }}
+  AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+  CLEANUP_AWS_ACCESS_KEY_ID: ${{ secrets.CLEANUP_AWS_ACCESS_KEY_ID }}
+  CLEANUP_AWS_SECRET_ACCESS_KEY: ${{ secrets.CLEANUP_AWS_SECRET_ACCESS_KEY }}
+  NSERVICEBUS_AMAZONSQS_S3BUCKET: ${{ secrets.NSERVICEBUS_AMAZONSQS_S3BUCKET }}
```
DavidBoike, line 18: I know this is already in master, but I'm torn about having this at the global level here. That puts these values in environment variables for the entirety of the workflow. It would seem to be more least-privilege to have this at the step level for only the tests……but then we'd have this block duplicated for both the linux and windows tests.

Thoughts @bording?

bording: I don't think duplicating them makes sense, so unless there's a specific problem you're worried about, I'd say it's fine.

DavidBoike (approved):

PR #819, opened 2021-05-18
fix OCE catching, Task.Run, etc.

adamralph:
The changes to `src/NServiceBus.Transport.SQS.Tests/Cleanup.cs` aren't strictly required (at least now). They are defensive, in case anyone adds a token to one of the calls.

### DIFF
--- a/src/NServiceBus.Transport.SQS.TransportTests/Sending_poison_messages.cs
+++ b/src/NServiceBus.Transport.SQS.TransportTests/Sending_poison_messages.cs
@@ -71,8 +71,10 @@
                 var messageReceived = false;
                 ReceiveMessageResponse receiveMessageResponse = null;
 
-                while (messageReceived == false && !cancellationToken.IsCancellationRequested)
+                while (messageReceived == false)
                 {
+                    cancellationToken.ThrowIfCancellationRequested();
+
                     receiveMessageResponse = await sqsClient.ReceiveMessageAsync(new ReceiveMessageRequest
                     {
                         QueueUrl = getQueueUrlResponse.QueueUrl,
--- a/src/NServiceBus.Transport.SQS.TransportTests/Sending_poison_messages.cs
+++ b/src/NServiceBus.Transport.SQS.TransportTests/Sending_poison_messages.cs
@@ -71,7 +71,7 @@
                 var messageReceived = false;
                 ReceiveMessageResponse receiveMessageResponse = null;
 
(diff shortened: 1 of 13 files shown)


src/NServiceBus.Transport.SQS/DelayedMessagesPump.cs
```diff
@@ -292,67 +299,85 @@ async Task SendDelayedMessagesInBatches(BatchEntry<SqsReceivedDelayedMessage> ba
...
                     {
                         var builder = new StringBuilder();
-                        foreach (var failed in changeVisibilityResult.Failed)
+                        foreach (var failed in result.Failed)
                         {
                             builder.AppendLine($"{failed.Id}: {failed.Message} | {failed.Code} | {failed.SenderFault}");
                         }
 
-                        Logger.Debug($"Changing visibility failed for {builder}");
+                        Logger.Error($"Error changing visibility for {builder}", ex);
```
andreasohlund, line 330: I assume this was changed to ERROR on purpose? any risk we're crying wolf in logs now?

andreasohlund (approved):

src/NServiceBus.Transport.SQS/DelayedMessagesPump.cs
adamralph: Yeah, it was on purpose. It seemed off to be logging an error at debug level. Anyone running at anything less than info would never know there is any problem here.

However, given the comment above says:

> // deliberately fire&forget because we treat this as a best effort

Perhaps this should be only a warning?

andreasohlund: WARN sounds like a good trade off

DavidBoike (approved):

src/NServiceBus.Transport.SQS.TransportTests/Sending_poison_messages.cs
```diff
@@ -71,8 +71,10 @@ async Task CheckErrorQueue(string errorQueueName, CancellationToken cancellation
                 var messageReceived = false;
                 ReceiveMessageResponse receiveMessageResponse = null;
 
-                while (messageReceived == false && !cancellationToken.IsCancellationRequested)
+                while (messageReceived == false)
```
andreasohlund, line 74:
```suggestion
                while (!messageReceived)
```

andreasohlund (approved):

Issue #866, opened 2021-06-18
Exception From Send (Rate Exceeded) 


Selimmo:
Hi Guys 
We are trying to convert some services to use the new **NserviceBuys.AmazonSQS 5.3.1** from **4.4.2**, those services subscribing on average to 40 topics each they are all deployed to one Ec2 instance, and we still in compatibility mode. 

I chose not to allow NserviceBus create the SNS policies and instead I made it myself to avoid any limitations problems with the AWS IAM policy size. 

Now the problem is we have been facing this exception a lot and it causes some failures with some of the events, we have some sagas that cause some Events to be dispatched in the same time which was working perfectly with the Message driven style but it is causing troubles with SQS/SNS style. 

All of that was discovered by some coverage tests that we had, again those are running fine with 4.4.2, but not with 5.3.1 and when we introduced some wait in the tests between those operations, it starts succeeding and i don't see the exception, but in a high load prod env, we can't control that. 

```
exceptionType
:
"Amazon.SimpleNotificationService.AmazonSimpleNotificationServiceException",
exceptionMessage
:
"Rate exceeded",
stackTrace
:
"   at Amazon.Runtime.Internal.HttpErrorResponseExceptionHandler.HandleException(IExecutionContext executionContext, HttpErrorResponseException exception)
   at Amazon.Runtime.Internal.ErrorHandler.ProcessException(IExecutionContext executionContext, Exception exception)
   at Amazon.Runtime.Internal.ErrorHandler.<InvokeAsync>d__5`1.MoveNext()
   at Amazon.Runtime.Internal.CallbackHandler.<InvokeAsync>d__9`1.MoveNext()
   at Amazon.Runtime.Internal.EndpointDiscoveryHandler.<InvokeAsync>d__2`1.MoveNext()
   at Amazon.Runtime.Internal.CredentialsRetriever.<InvokeAsync>d__7`1.MoveNext()
   at Amazon.Runtime.Internal.RetryHandler.<InvokeAsync>d__10`1.MoveNext()
   at Amazon.Runtime.Internal.CallbackHandler.<InvokeAsync>d__9`1.MoveNext()
   at Amazon.Runtime.Internal.ErrorCallbackHandler.<InvokeAsync>d__5`1.MoveNext()
   at Amazon.Runtime.Internal.MetricsHandler.<InvokeAsync>d__1`1.MoveNext()
[...]

PR #879, opened 2021-06-29
866: removed the code that tries to list the topics with each dispatc…

Selimmo:
Because of the issue https://github.com/Particular/NServiceBus.AmazonSQS/issues/866

The code that is removed (line 247 to 256) was causing many failures, it is causing each dispatch to go and `ListTopics` which has a 30 transactions per second Hard limit.

we first downloaded the code to our repo and used it instead of `5.3.1 nuget package`, and all the tests started to succeed. 

was hoping to speed this up a little by creating the PR (we can maybe create a new `release 5.4`)

### DIFF
--- a/src/NServiceBus.Transport.SQS/MessageDispatcher.cs
+++ b/src/NServiceBus.Transport.SQS/MessageDispatcher.cs
@@ -12,7 +12,6 @@
     using Amazon.SQS.Model;
     using DelayedDelivery;
     using Extensibility;
-    using Extensions;
     using Logging;
     using SimpleJson;
     using Transport;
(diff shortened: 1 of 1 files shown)


ramonsmits:
@Selimmo The code removed the lookup but from what I understood it existed to detect incorrectly configured SQS configurations. This PR would make sense if we would drop prior configurations but that would be a breaking change and would require a new major version.

The team what will resolve the issue once prioritized and staffed will definitely look at this PR but we can't merge this as is as a new minor or major.

Selimmo:
Thanks @ramonsmits 

I was wondering if you or someone could give some more details on "incorrectly configured SQS configurations"? maybe an example of such incorrect configuration would be great.

we can also modify this code to Opt-out so the developer (i.e. us) can choose not to have this function if we are confident that it will not be incorrectly configured in our environment?

we are in this situation because of AWS hard limit with ListTopics and due to that i feel the current implementation is not Fit for purpose.

Thanks 
Mo

ramonsmits:
> I was wondering if you or someone could give some more details on "incorrectly configured SQS configurations"? maybe an example of such incorrect configuration would be great.

@Selimmo It's connected to the message-driven pub/sub hybrid mode. If the message we're trying to dispatch is a unicast message with a Publish intent but the subscriber is also subscribed via SNS we don't want to dispatch the message twice, the subscriber will receive it via SNS and not via a unicast send.

PR #964, opened 2021-08-27
TF 753 acceptance tests cleanup scheduled lambda

cquirosj:
[TF work](https://github.com/Particular/KeepTheLightsOn/issues/753)

This PR Includes the lambda function code itself along with the code that takes care of the deployment of the lambda function with all its required artifacts like Roles, RolePolicies, and EventBridge Rule (CloudWatch Event) configuration.

Still pending:
- [x] Making code analysis checks pass.
- [ ] Convert hardcoded schedule frequency into a parameter in the pulumi stack YAML file.
- [ ] Update the documentation about the existence of the scheduled task and how to execute its deployment if the lambda function code ever gets updated. 
- [ ] Define the best folder name for the lambda scheduled function and its deployment "script."
- [ ] Nice to have: define if we want to add SNS notifications when the lambda function fails.
- [ ] Discuss naming conventions for the 7 cloud objects created by the deployment "script."
- [ ] Discuss possible code Reutilization: since the deployment "script" is written in C# (infrastructure as software), there is a chance to make the lambda function re-utilize the same code of https://github.com/Particular/NServiceBus.AmazonSQS/blob/master/src/NServiceBus.Transport.SQS.Tests/Cleanup.cs instead of duplicating it
- [ ] Discuss whether or not we want the Lambda function to delete not only S3 buckets but also delete any subscriptions, topics, and queues potentially being left behind too.

### DIFF
--- a/src/Tests.DeleteS3BucketsScheduledLambda/deployment/DeleteS3BucketsStack.cs
+++ b/src/Tests.DeleteS3BucketsScheduledLambda/deployment/DeleteS3BucketsStack.cs
@@ -4,9 +4,9 @@ using Pulumi.Aws.Lambda;
 using Pulumi.Aws.CloudWatch;
 
 
-internal class DeleteS3BucketsStack : Stack
+class DeleteS3BucketsStack : Stack
 {
-    private const int LambdaMaxAllowedExecutionTimeInSeconds = 60;
+    const int LambdaMaxAllowedExecutionTimeInSeconds = 60;
     public DeleteS3BucketsStack()
     {
         var lambda = new Function("DeleteSQSTransportTestsExpiredAssets",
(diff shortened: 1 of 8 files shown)


DavidBoike:
Looking at this implementation I’m actually a bit concerned about the complexity and number of dependencies involved.

In contrast, GitHub Actions runners have pwsh, already have the [AWS CLI preinstalled](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#cli-tools), can run on a cron schedule, and already have capabilities built in to GitHub to see and diagnose failures.

cquirosj:
> In contrast, GitHub Actions runners have pwsh, already have the [AWS CLI preinstalled](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#tools), can run on a cron schedule, and already have capabilities built in to GitHub to see and diagnose failures.

GitHub Actions runners also already come with pulumi pre-installed. The only remaining aspect is to define what we want to use for diagnosing. However, I totally agree with can do this with pure GitHub actions only executing a pwsh script. Only more fun for me if I get to write that plan too :)

Let me demo when you have time how the solution with pulumi works and we can decide if we go with pure github actions.

cquirosj:
Closed in favor of https://github.com/Particular/NServiceBus.AmazonSQS/pull/988

PR #1001, opened 2021-09-16
Update project target framework

HEskandari:


src/CommandLine/NServiceBus.Transports.SQS.CommandLine.csproj
```diff
@@ -1,9 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
     <OutputType>Exe</OutputType>
     <GenerateDocumentationFile>false</GenerateDocumentationFile>
+    <LangVersion>8.0</LangVersion>
```
bording, line 7: C# 8 is the default for `netcoreapp3.1`, so technically we don't need this for any projects that are are single targeting just `netcoreapp3.1`. 

I guess we can just add to every project regardless, but I'm kind of inclined to only add it where it's needed: multi-targeted projects that include a .NET Framework TFM.

HEskandari: I thought to add it for consistency, but we can just add it where needed. I'll update.

bording (approved):

Issue #1158, opened 2021-11-26
SQS visiblity period not using custom default value


fadilafidina:
Hi,

We are using NServiceBus with SQS, without compatibility mode.

We think that all of our messages are defaulting to have a 30s visibility period timeout, even though we have set up a timeout of 2 mins 30 seconds from the AWS console itself for our queue.

We have a message handled in a saga that takes around 60-90 seconds to process, and we are seeing it get processed multiple times. We think it's because of the visibility period defaulting to 30 seconds, therefore the message was visible again on the queue, hence the multiple processing of the same message. However, we expect it to be using the 2 minutes 30 seconds setting that we have configured.

We see the same message being handled multiple times in quick succession, 0.1 of a second apart.

Our saga is set up for pessimistic locking.

Is there a reason why it's not adhering to the viability timeout we've configured?

We've also read that in https://docs.particular.net/nservicebus/sagas/concurrency#high-load-scenarios-avoid-io-or-cpu-bound-operations-in-a-saga-handler, we should not be processing long operations in a saga handler. Data contention is not a concern at the time that this process is executed.

mauroservienti:
@fadilafidina thanks for raising this. I transferred the issue to the appropriate repository.

What version of the transport are you using?

You are correct, we are not honoring any message visibility timeout value defined on the main input queue. In the meantime, a workaround that might work for you is to enable the [message-driven pub/sub compatibility](https://docs.particular.net/transports/sqs/configuration-options#message-driven-pubsub-compatibility-mode) mode (via the `EnableMessageDrivenPubSubCompatibilityMode` config option) even if you don't need it, and set the message visibility timeout there:

```
var migrationSettings = transport.EnableMessageDrivenPubSubCompatibilityMode();
migrationSettings.MessageVisibilityTimeout(timeoutInSeconds: 300);
``` 

There might be a little performance penalty at publishing time if the endpoint publishes a lot of events. However, the built-in topics and subscriptions cache should compensate for that.

---

@danielmarbach this sounds like a nice-to-have improvement. IIRC we already retrieve information from the queues adding one more attribute should not be a big deal.

fadilafidina:
Thanks for your prompt reply @mauroservienti! And thanks for moving the issue across - I realised afterwards that it should move across. 

We are using 5.4.0. 

We just removed compatibility mode from our code and have deployed it across most of our services, with the last bit of it being very soon. It would be great if we don't have to add it back in since it's being deprecated.

[...]

PR #1159, opened 2021-11-28
Set ReceiveMessageRequest VisibilityTimeout only if configured

mauroservienti:
Related to https://github.com/Particular/NServiceBus.AmazonSQS/issues/1158

### DIFF
--- a/src/NServiceBus.Transport.SQS/InputQueuePump.cs
+++ b/src/NServiceBus.Transport.SQS/InputQueuePump.cs
@@ -85,8 +85,12 @@ namespace NServiceBus.Transport.SQS
                 WaitTimeSeconds = 20,
                 AttributeNames = new List<string> { "SentTimestamp" },
                 MessageAttributeNames = new List<string> { "*" },
-                VisibilityTimeout = configuration.MessageVisibilityTimeout,
             };
+
+            if (configuration.MessageVisibilityTimeout.HasValue)
+            {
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -85,9 +85,13 @@ public void Start(int maximumProcessingConcurrency, CancellationToken token)
                 WaitTimeSeconds = 20,
                 AttributeNames = new List<string> { "SentTimestamp" },
                 MessageAttributeNames = new List<string> { "*" },
-                VisibilityTimeout = configuration.MessageVisibilityTimeout,
             };
 
+            if (configuration.MessageVisibilityTimeout.HasValue)
```
mauroservienti, line 90: As discussed with @danielmarbach, it's important to keep in mind that this check might be misleading as it doesn't express the fact that the `MessageVisibilityTimeout` property is only set when using the pub/sub compatibility mode. It might be worth either documenting with a comment why the if exists or adding an additional condition like (pseudo-code):

```
if (compatibilityModeIsEnabled && configuration.MessageVisibilityTimeout.HasValue)
```

danielmarbach: I would prefer having a comment here. Otherwise good to go

danielmarbach (approved):

mauroservienti:
Replaced by #1235

PR #1204, opened 2022-01-11
Update AWS SDK dependencies to 3.7.*

kbaley:
Updated the upper bound of the AWS SDKs to match master. They are capped because the SDKs don't always follow SemVer.


DavidBoike:
Should update the test projects to the latest 3.7 versions as well, right?

bording (approved):

PR #1205, opened 2022-01-12
Update aws dependencies for v5.4

kbaley:
Updated the upper bound of the AWS SDKs to match master. They are capped because the SDKs don't always follow SemVer.


kbaley:
Closing. Updated SDK is only to be done on the latest supported version.

PR #1211, opened 2022-01-13
Fix AWS dependency ranges

kbaley:
Minimum dependency number should be more specific for the transport project since nuget defaults to pulling the lowest version number that meets the range.

I also bumped the version numbers of the AWS SDK packages in the test projects to match those used in the main project


bording (changes requested):
Given that NuGet packages are "supposed" to only have 3 version components, let's go with `3.8.0` for the upper bound.

Otherwise LGTM

bording (approved):

PR #1237, opened 2022-02-07
[5.4.1] Respect the message visibility timeout defined at the queue level

soujay:
In [NServiceBus.AmazonSQS 5.4](https://github.com/Particular/NServiceBus.AmazonSQS/releases/tag/5.4.0) a rate limiter and in-memory cache were introduced to fix a "rate exceeded" exception when using pub/sub compatibility mode—introducing the fix required to allow users to explicitly set the message visibility timeout when using the compatibility mode.

Unfortunately, due to the way it was implemented a message visibility timeout value is always set even if the compatibility mode is not enabled. When compatibility mode is not enabled the default value of 30 seconds is used for the visibility timeout and this causes the ReceiveMessageRequest to not use any visibility timeout specified on the queue when creating it.

This PR changes the transport behavior and stops setting a visibility timeout value if not explicitly set by the user code.

Related to #1158

### DIFF
--- a/src/NServiceBus.Transport.SQS/Configure/SqsSubscriptionMigrationModeSettings.cs
+++ b/src/NServiceBus.Transport.SQS/Configure/SqsSubscriptionMigrationModeSettings.cs
@@ -46,7 +46,7 @@
         }
 
         /// <summary>
-        /// Overrides the default value of 30 seconds for SQS message visibility timeout.
+        /// Overrides the default value specified at the queue level for message visibility timeout.
         /// </summary>
         /// <param name="timeoutInSeconds">Message visibility timeout.</param>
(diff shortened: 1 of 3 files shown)


bording:
I think the xml comment for setting also needs to be updated: https://github.com/Particular/NServiceBus.AmazonSQS/blob/release-5.6/src/NServiceBus.Transport.SQS/Configure/SqsSubscriptionMigrationModeSettings.cs#L49

A queue has a default timeout of 30 seconds, but since that can be changed at the queue level, that xml comment to indicate that using that API is overriding the queue's configured timeout, with no mention of it being 30 seconds since we have no way of knowing what the queue's actual timeout value is.

soujay:
@bording - xml comment has been updated. Please review once more.

bording (approved):

PR #1535, opened 2022-08-15
update ci.yaml

awright18:
following procedures from https://github.com/Particular/KeepTheLightsOn/issues/1103


DavidBoike (approved):
Change looks fine to me. No idea why the failure though.

PR #1597, opened 2022-09-20
Bump AWS SDK dependencies

mauroservienti:
Bump dependencies all at once in an attempt to overcome the chicken and egg problem caused by https://github.com/aws/aws-sdk-net/issues/1998


danielmarbach:
@DavidBoike yeah we could definitely. It was just something that I have us not seeing a lot and frankly speaking, I was glad since I joined Particular I never had to deal with things like Moq, FakeItEasy and friends again since they come with their own quirky syntax that you have to learn, understand...

Anyway, just a brief history of why we didn't do it here. We have similar cases in other repos where we create these manual fakes (and many of them are probably also my fault :grinning: )

danielmarbach (approved):

PR #1643, opened 2022-10-05
Pre-deploy infrastructure for hybrid Pub/Sub tests and ensure queue name uniqueness

mauroservienti:
## Problem description

AWS is an eventually consistent environment and requires up to a `60` seconds delay until an SNS access policy is fully applied. In our case, the policy defines how events get forwarded from topics to queues. The delay is likely caused by the fact that there are multiple nodes provisioned for serving any given topic/queue, and the policy has to propagate to all of them. This shouldn't be an issue for most production environments. However, for the hybrid pub-sub tests where each test case deploys a new infrastructure and sends a non-trivial number of messages, it resulted in failing CI builds. The tests were failing with timeout exceptions due to the policy still being propagated while the tests were already running. Multiple re-runs of the failed tests were needed in order for it to pass.  

Secondly, the queue name generation strategy caused queue name clashes. The name was derived from a string `{TestName}_{EndpointName}` that was truncated at the beginning to keep the length under the maximal allowed value.  This resulted in tests sharing queue names and in some cases resulted in subscription messages being picked up by the wrong test. When that happened, the test waiting for a subscription message would never get one and eventually fail with a timeout exception.

With each run of the CI taking roughly 40-60 mins, this was becoming a very tedious and wasteful development process.

## Implemented solution

This PR introduces a pre-deploy infrastructure step for tests that publish more than one event and have more than one test case. The test case execution is delayed for 60 seconds to ensure that AWS access policy has fully propagated. 

There were roughly 20 test cases changed which caused the CI run time to increase but ~20 min. As a result, a decision was made,  to keep only the smallest and the largest test cases. Others have been commented out but not completely removed as it still might be valuable to run these test cases manually e.g. during releases.

In addition, a per-test-case name prefix is applied to avoid queue name collision.


mauroservienti:
From [documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request):

> Note: Workflows will not run on pull_request activity if the pull request has a merge conflict. The merge conflict must be resolved first.

🤷‍♂️🤦🏼‍♂️

mauroservienti:
@TravisNickels @tmasternak, I rebased and force-pushed. Actions are running again

mauroservienti:
🎉

src/NServiceBus.Transport.SQS.AcceptanceTests/NativePubSub/When_concurrent_subscribe.cs
```diff
@@ -15,6 +15,7 @@ public async Task It_should_still_deliver()
                 .WithEndpoint<Publisher>(b => b.When(c => c.Subscribed, (session, ctx) => Task.WhenAll(session.Publish(new MyEvent()), session.Publish(new MyOtherEvent()))))
                 .WithEndpoint<Subscriber>(b => b.When(async (session, ctx) =>
                 {
+                    // TODO: this part is the culprit
```
mauroservienti, line 18: Revert this before merging

src/NServiceBus.Transport.SQS.AcceptanceTests/NativePubSub/HybridModeRateLimit/When_publishing_two_event_types_to_native_and_non_native_subscribers_in_a_loop_in_the_context_of_incoming_message.cs
```diff
@@ -270,5 +311,17 @@ public class MyEvent : IEvent
...
+        {
+            static ILog log = LogManager.GetLogger<IncomingLoggingBehavior>();
+
+            public override async Task Invoke(IIncomingPhysicalMessageContext context, Func<Task> next)
+            {
+                log.Debug($"-----> {context.Message.MessageId}");
+
+                await next();
+            }
+        }
```
mauroservienti, line 325: Revert this before merging

tmasternak (approved):

PR #1692, opened 2022-10-29
Update the max backoff interval to 60 seconds

mauroservienti:
Travis, Tomek, it seems to me that it makes sense to align the maximum backoff interval for command line tests to the overall value used by AWS for changes propagation.

For example, one of the command line flaky tests is the one that removes S3 buckets and included resources, the test waits for a certain amount of time before failing. That amount of time is 20 seconds, this PR changes that to 60

Thoughts?


mauroservienti:
Uhm, no. I raised https://github.com/Particular/NServiceBus.AmazonSQS/issues/1693

tmasternak (approved):

TravisNickels (approved):

PR #1861, opened 2023-01-19
Return message to queue between retry attempts

awright18:


src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -168,6 +169,7 @@ public async Task ChangeConcurrency(PushRuntimeSettings limitations, Cancellatio
             await StartReceive(cancellationToken).ConfigureAwait(false);
         }
 
+
```
bording, line 172: :fire:

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -288,6 +293,11 @@ internal async Task ProcessMessage(Message receivedMessage, CancellationToken me
                 isPoisonMessage = true;
             }
 
+            if (messagesToBeDeleted.TryGet(messageId, out _))
```
bording, line 296: We should move this check to directly after the `messageId` value is assigned. We don't care about the `s3BodyKey` value at this point because if the message is in `messagesToBeDeleted` we've already called `DeleteMessage` once and logged the message if the body is in S3.

bording: Also, we should `return` if we find this message because we want to skip all additional processing as this point.

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -308,69 +318,113 @@ internal async Task ProcessMessage(Message receivedMessage, CancellationToken me
...
+            }
+        }
+
 
-            // Always delete the message from the queue.
-            // If processing failed, the onError handler will have moved the message
-            // to a retry queue.
-            await DeleteMessage(receivedMessage, transportMessage.S3BodyKey, messageProcessingCancellationToken).ConfigureAwait(false);
+#pragma warning disable PS0018 //Cancellation token intentionally not passed because cancellation shouldn't stop messages from being returned to the queue
+        async Task ReturnMessageToQueue(Message message)
```
bording, line 331: Let's move this down below with the others as well.

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -308,69 +318,113 @@ internal async Task ProcessMessage(Message receivedMessage, CancellationToken me
...
-                // set the native message on the context for advanced usage scenario's
-                var context = new ContextBag();
-                context.Set(nativeMessage);
-                // We add it to the transport transaction to make it available in dispatching scenario's so we copy over message attributes when moving messages to the error/audit queue
-                var transportTransaction = new TransportTransaction();
-                transportTransaction.Set(nativeMessage);
-                transportTransaction.Set("IncomingMessageId", headers[Headers.MessageId]);
+                var deliveryAttempts = GetDeliveryAttempts(nativeMessageId);
+
+                var errorHandlerResult = ErrorHandleResult.RetryRequired;
```
bording, line 385: Not seeing a reason to pre-declare this vs on line 389

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -405,13 +459,21 @@ async Task DeleteMessage(Message message, string s3BodyKey, CancellationToken me
...
             {
                 Logger.Info($"Message receipt handle {message.ReceiptHandle} no longer valid.", ex);
                 return; // if another receiver fetches the data from S3
             }
+            catch (Exception ex) when (!ex.IsCausedBy(messageProcessingCancellationToken))
+            {
+                Logger.Info($"Failed to delete message with receipt handle {message.MessageId}");
+            }
+
+            messagesToBeDeleted.AddOrUpdate(message.MessageId, true);
```
bording, line 476: We only need to do this if `DeleteMessageAsync` fails, so it should be inside a `catch`

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -473,6 +535,7 @@ async Task MovePoisonMessageToErrorQueue(Message message, CancellationToken mess
             // If there is a message body in S3, simply leave it there
         }
 
+
```
bording, line 538: :fire:

src/NServiceBus.Transport.SQS/NServiceBus.Transport.SQS.csproj
```diff
@@ -16,6 +16,7 @@
     <PackageReference Include="AWSSDK.S3" Version="[3.7.101.10, 3.8.0)" />
     <PackageReference Include="AWSSDK.SimpleNotificationService" Version="[3.7.100.10, 3.8.0)" />
     <PackageReference Include="AWSSDK.SQS" Version="[3.7.100.9, 3.8.0)" />
+    <PackageReference Include="BitFaster.Caching" Version="[2.0.0, 3.0.0)" />
```
bording, line 19:
```suggestion
    <PackageReference Include="BitFaster.Caching" Version="[2.1.1, 3.0.0)" />
```

bording (changes requested):

src/NServiceBus.Transport.SQS/InputQueuePump.cs
awright18: delete the empty line

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -308,69 +317,111 @@ internal async Task ProcessMessage(Message receivedMessage, CancellationToken me
...
+                Logger.Warn("Failed to return message to the queue. The message is not in flight.", ex);
+            }
+            catch (Exception ex)
+            {
+                Logger.Warn(
+                    "Failed to return message to the queue. Message will return to queue after visibility expires.", ex);
+            }
+        }
+
+        int GetDeliveryAttempts(string messageIdKey)
```
bording, line 418: Let's rename `messageIdKey` to `nativeMessageId` to make it clearer what needs to be passed in.

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -308,69 +317,111 @@ internal async Task ProcessMessage(Message receivedMessage, CancellationToken me
...
+            {
+                await sqsClient.ChangeMessageVisibilityAsync(
+                    new ChangeMessageVisibilityRequest()
+                    {
+                        QueueUrl = inputQueueUrl,
+                        ReceiptHandle = message.ReceiptHandle,
+                        VisibilityTimeout = 0
+                    }).ConfigureAwait(false);
+            }
+            catch (ReceiptHandleIsInvalidException ex)
```
bording, line 403: Not sure these specialized `catch` blocks are really adding anything here. Let's simplify this down to just the `Exception` one.

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -405,12 +456,14 @@ async Task DeleteMessage(Message message, string s3BodyKey, CancellationToken me
             try
             {
                 // should not be canceled
-                await sqsClient.DeleteMessageAsync(inputQueueUrl, message.ReceiptHandle, messageProcessingCancellationToken).ConfigureAwait(false);
+                await sqsClient
+                    .DeleteMessageAsync(inputQueueUrl, message.ReceiptHandle, messageProcessingCancellationToken)
+                    .ConfigureAwait(false);
             }
-            catch (ReceiptHandleIsInvalidException ex)
+            catch (Exception ex) when (!ex.IsCausedBy(messageProcessingCancellationToken))
```
bording, line 463: I think we want to keep this as a `ReceiptHandleIsInvalidException` `catch`, but tweak the wording and the `messagesToBeDeleted` stuff.

Looks like we should also move the `s3BodyKey` `if` block to be inside the inside the `try`.

bording (changes requested):

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -308,69 +317,103 @@ internal async Task ProcessMessage(Message receivedMessage, CancellationToken me
...
             }
-
-            // Always delete the message from the queue.
-            // If processing failed, the onError handler will have moved the message
-            // to a retry queue.
-            await DeleteMessage(receivedMessage, transportMessage.S3BodyKey, messageProcessingCancellationToken).ConfigureAwait(false);
         }
 
-        async Task ProcessMessageWithInMemoryRetries(Dictionary<string, string> headers, string nativeMessageId, byte[] body, Message nativeMessage, CancellationToken messageProcessingCancellationToken)
+        async Task InnerProcessMessage(Dictionary<string, string> headers, string nativeMessageId, byte[] body, string s3BodyKey, Message nativeMessage, CancellationToken messageProcessingCancellationToken)
```
bording, line 328: I'm thinking we could change this method to return a `bool` that this true if the message was processed successfully instead of deleting the message in this method. That would let us remove having to pass in the `s3BodyKey.

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -308,69 +317,103 @@ internal async Task ProcessMessage(Message receivedMessage, CancellationToken me
             if (!IsMessageExpired(receivedMessage, transportMessage.Headers, messageId, CorrectClockSkew.GetClockCorrectionForEndpoint(awsEndpointUrl)))
```
bording, line 317: Let's invert this condition, so it becomes "if expired, delete, else process"

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -308,69 +317,103 @@ internal async Task ProcessMessage(Message receivedMessage, CancellationToken me
...
             {
                 // here we also want to use the native message id because the core demands it like that
-                await ProcessMessageWithInMemoryRetries(transportMessage.Headers, nativeMessageId, messageBody, receivedMessage, messageProcessingCancellationToken).ConfigureAwait(false);
+                await InnerProcessMessage(transportMessage.Headers, nativeMessageId, messageBody, transportMessage.S3BodyKey, receivedMessage, messageProcessingCancellationToken).ConfigureAwait(false);
```
bording, line 320: We can use the `bool` I proposed to return to check if we should call `DeleteMessage` or not

bording (changes requested):

bording (approved):

PR #1882, opened 2023-02-01
TestQueueDoesNotExistErrorMessage

jpalac:
Ensure that the exception thrown when sending to a queue that does not exist includes the queue name.


mikeminutillo (approved):

mikeminutillo:
This should already pass on master but once it's in we'll backport it to the release branches. We expect that it will fail for v5 and then we'll fix it there

danielmarbach:
LGTM

PR #1885, opened 2023-02-01
Replace SimpleJson with System.Text.Json

### LINKED ISSUE #1213: Replace SimpleJson
The SimpleJson codebase is not maintained and is fairly old. This constitutes a maintenance risk, and while there are no known vulnerabilities, if there ever were any detected, it's unlikely there would be a fix.

It should be possible to replace SimpleJson, perhaps with System.Text.Json, although some migration may be required.

danielmarbach:
Fixes #1213 

- [x] Add acceptance test with fixed message content as string serialized with system text json and verify if it can be received
- [x] Smoke test things with previous releases including V1 version?


src/NServiceBus.Transport.SQS.AcceptanceTests/Sending_in_compatibility_mode.cs
```diff
@@ -25,46 +25,37 @@ public class Context : ScenarioContext
...
         public class Sender : EndpointConfigurationBuilder
         {
-            public Sender()
-            {
+            public Sender() =>
                 EndpointSetup<DefaultServer>(builder =>
                 {
                     builder.ConfigureRouting().RouteToEndpoint(typeof(Message), typeof(Receiver));
-                    builder.ConfigureSqsTransport();
+                    builder.ConfigureSqsTransport().EnableV1CompatibilityMode = true;
```
danielmarbach, line 32: This is weird because apparently while migrating we have essentially removed this flag

https://github.com/Particular/NServiceBus.AmazonSQS/blob/release-5.6/src/NServiceBus.Transport.SQS.AcceptanceTests/Sending_in_compatibility_mode.cs#L33

src/NServiceBus.Transport.SQS/ReducedPayloadSerializerConverter.cs
```diff
@@ -0,0 +1,41 @@
...
+        static readonly JsonSerializerOptions TransportMessageSerializerOptions = new()
+        {
+            TypeInfoResolver = TransportMessageSerializerContext.Default
+        };
+
+        static readonly JsonConverter<Dictionary<string, string>> DefaultDictionaryConverter =
+            (JsonConverter<Dictionary<string, string>>)JsonSerializerOptions.Default.GetConverter(typeof(Dictionary<string, string>));
+        public override TransportMessage Read(ref Utf8JsonReader reader, Type typeToConvert,
+            JsonSerializerOptions options) =>
+            JsonSerializer.Deserialize<TransportMessage>(ref reader, TransportMessageSerializerOptions);
```
danielmarbach, line 25: In practice, we never use the read method. Should I throw here instead?

src/NServiceBus.Transport.SQS/NServiceBus.Transport.SQS.csproj
```diff
@@ -16,10 +15,11 @@
     <PackageReference Include="AWSSDK.S3" Version="[3.7.101.10, 3.8.0)" />
     <PackageReference Include="AWSSDK.SimpleNotificationService" Version="[3.7.100.10, 3.8.0)" />
     <PackageReference Include="AWSSDK.SQS" Version="[3.7.100.9, 3.8.0)" />
-    <PackageReference Include="NServiceBus" Version="[8.0.0, 9)" />
+    <PackageReference Include="NServiceBus" Version="[8.0.0, 9.0.0)" />
     <PackageReference Include="Fody" Version="6.6.4" PrivateAssets="All" />
     <PackageReference Include="Obsolete.Fody" Version="5.3.0" PrivateAssets="All" />
     <PackageReference Include="Particular.Packaging" Version="2.3.0" PrivateAssets="All" />
+    <PackageReference Include="System.Text.Json" Version="7.0.1" />
```
danielmarbach, line 22: Initially, this one was for NET472 but we had to move it because otherwise the `.Default` property is not available.

src/NServiceBus.Transport.SQS/ReducedPayloadSerializerConverter.cs
```diff
@@ -0,0 +1,41 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Text.Json;
+    using System.Text.Json.Serialization;
+
+    /// <summary>
+    /// This serializer is only used to make sure we are not double storing TimeToBeReceived and ReplyAddress both in
+    /// the headers and on the transport message wrapper to save space on the SQS message body.
+    /// The read method is implemented but never used because only the dispatcher uses it to serialize the transport
+    /// message.
```
danielmarbach, line 12:
```suggestion
    /// This converter is only used to make sure we are not double storing TimeToBeReceived and ReplyAddress both in
    /// the headers and on the transport message wrapper to save space on the SQS message body.
    /// The read method is implemented but never used because only the dispatcher uses it to serialize the transport
    /// message.
```

src/NServiceBus.Transport.SQS/ReducedPayloadSerializerConverter.cs
```diff
@@ -0,0 +1,41 @@
...
+        public override TransportMessage Read(ref Utf8JsonReader reader, Type typeToConvert,
+            JsonSerializerOptions options) =>
+            JsonSerializer.Deserialize<TransportMessage>(ref reader, TransportMessageSerializerOptions);
+
+        public override void Write(Utf8JsonWriter writer, TransportMessage value, JsonSerializerOptions options)
+        {
+            writer.WriteStartObject();
+
+            writer.WriteString(nameof(TransportMessage.Body), value.Body);
+            writer.WriteString(nameof(TransportMessage.S3BodyKey), value.S3BodyKey);
```
mikeminutillo, line 32: Should we consider adding an approval test for the properties on TransportMessage?
If someone adds one in the future, it will cause a subtle error that may be difficult to pick up.

src/NServiceBus.Transport.SQS/ReducedPayloadSerializerConverter.cs
mikeminutillo: I think so. It's better not to have code we don't need. It can be restored later if we need it for whatever reason.

mikeminutillo (approved):

src/NServiceBus.Transport.SQS.AcceptanceTests/Sending_in_compatibility_mode.cs
mikeminutillo: That stood out me as well. Is this test working as expected?

mikeminutillo:
> Mauro and I discussed that it would probably be helpful to start obsoleting with warn the V1Compatibility support. Any thoughts on that?

I think that makes sense. It's been out of support since Oct 2018, which is over 4 years ago.

src/NServiceBus.Transport.SQS/ReducedPayloadSerializerConverter.cs
danielmarbach: Found a way to cleanup existing tests into approvals.

src/NServiceBus.Transport.SQS.AcceptanceTests/Sending_in_compatibility_mode.cs
danielmarbach: I improved the test and added another one to make this better

mauroservienti (approved):

jpalac (approved):

danielmarbach:
Smoke testing completed

Issue #1249, opened 2022-02-11
Empty message bodies cause deserialization exception


kentdr:
When messages are sent that have an empty message body, the transport [sets the message body to the literal "empty message". ](https://github.com/Particular/NServiceBus.AmazonSQS/blob/96d7054075c47318ad75de8fc26ec8a26dc4dcac/src/NServiceBus.Transport.SQS/TransportMessage.cs#L33)

When these messages are received, the "empty message" case is not accounted for and the message is treated as if it were a[ normal message with a Base64 string as the body.](https://github.com/Particular/NServiceBus.AmazonSQS/blob/96d7054075c47318ad75de8fc26ec8a26dc4dcac/src/NServiceBus.Transport.SQS/Extensions/MessageExtensions.cs#L18) .    In this case, the RetrieveBody method returns garbage text (a decoded base64 string that was never base64'd) causing message deserialization to fail.

PR #1889, opened 2023-02-02
Empty message body fix

jpalac:
Fix https://github.com/Particular/NServiceBus.AmazonSQS/issues/1249

### DIFF
--- a/src/NServiceBus.Transport.SQS.Tests/InputQueuePumpTests.cs
+++ b/src/NServiceBus.Transport.SQS.Tests/InputQueuePumpTests.cs
@@ -116,7 +116,7 @@ namespace NServiceBus.Transport.SQS.Tests
                     {Headers.MessageId, messageId},
                     {TransportHeaders.TimeToBeReceived, ttbr.ToString()}
                 },
-                Body = null
+                Body = string.Empty
             });
 
             var message = new Message
@@ -161,7 +161,7 @@ namespace NServiceBus.Transport.SQS.Tests
                 {
                     {Headers.MessageId, messageId}
                 },
-                Body = null
+                Body = string.Empty
             });
 
             var message = new Message
--- a/src/NServiceBus.Transport.SQS.Tests/InputQueuePumpTests.cs
+++ b/src/NServiceBus.Transport.SQS.Tests/InputQueuePumpTests.cs
@@ -116,7 +116,7 @@ namespace NServiceBus.Transport.SQS.Tests
                     {Headers.MessageId, messageId},
                     {TransportHeaders.TimeToBeReceived, ttbr.ToString()}
                 },
-                Body = "empty message"
+                Body = TransportMessage.EmptyMessage
             });
 
             var message = new Message
@@ -161,7 +161,7 @@ namespace NServiceBus.Transport.SQS.Tests
                 {
                     {Headers.MessageId, messageId}
                 },
-                Body = "empty message"
+                Body = TransportMessage.EmptyMessage
             });
 
             var message = new Message
--- a/src/NServiceBus.Transport.SQS.Tests/InputQueuePumpTests.cs
+++ b/src/NServiceBus.Transport.SQS.Tests/InputQueuePumpTests.cs
@@ -116,7 +116,7 @@ namespace NServiceBus.Transport.SQS.Tests
                     {Headers.MessageId, messageId},
                     {TransportHeaders.TimeToBeReceived, ttbr.ToString()}
                 },
-                Body = string.Empty
+                Body = TransportMessage.EmptyMessage
             });
 
             var message = new Message
@@ -161,7 +161,7 @@ namespace NServiceBus.Transport.SQS.Tests
                 {
                     {Headers.MessageId, messageId}
                 },
-                Body = string.Empty
+                Body = TransportMessage.EmptyMessage
             });
 
             var message = new Message
--- a/src/NServiceBus.Transport.SQS.Tests/InputQueuePumpTests.cs
+++ b/src/NServiceBus.Transport.SQS.Tests/InputQueuePumpTests.cs
@@ -116,7 +116,7 @@ namespace NServiceBus.Transport.SQS.Tests
                     {Headers.MessageId, messageId},
                     {TransportHeaders.TimeToBeReceived, ttbr.ToString()}
                 },
-                Body = TransportMessage.EmptyMessage
+                Body = null
             });
 
             var message = new Message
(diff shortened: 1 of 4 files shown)


src/NServiceBus.Transport.SQS/Extensions/MessageExtensions.cs
```diff
@@ -15,7 +15,14 @@ static class MessageExtensions
         {
             if (string.IsNullOrEmpty(transportMessage.S3BodyKey))
             {
-                return Convert.FromBase64String(transportMessage.Body);
+                if (transportMessage.Body == TransportMessage.EmptyMessage)
+                {
+                    return System.Text.Encoding.Default.GetBytes("");
```
mauroservienti, line 20: As we discussed during the sync, we should evaluate if this needs to be `UTF8` instead of `Default`

src/NServiceBus.Transport.SQS/TransportMessage.cs
```diff
@@ -28,9 +29,9 @@ public TransportMessage(OutgoingMessage outgoingMessage, DispatchProperties prop
             }
 
 #if NETFRAMEWORK
-            Body = outgoingMessage.Body.Length != 0 ? Convert.ToBase64String(outgoingMessage.Body.ToArray()) : "empty message";
+            Body = outgoingMessage.Body.Length != 0 ? Convert.ToBase64String(outgoingMessage.Body.ToArray()) : EmptyMessage;
 #else
-            Body = outgoingMessage.Body.Length != 0 ? Convert.ToBase64String(outgoingMessage.Body.Span) : "empty message";
+            Body = outgoingMessage.Body.Length != 0 ? Convert.ToBase64String(outgoingMessage.Body.Span) : EmptyMessage;
 #endif
```
mauroservienti, line 35: In the case of native integration, is there a chance that `Body` is `null`, or will it always be an empty string??

src/NServiceBus.Transport.SQS/Extensions/MessageExtensions.cs
danielmarbach: I'm not entirely sure if we have to use UTF8 or if we can use `Unicode`

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html

links to https://www.w3.org/TR/REC-xml/#charsets

which says

> The mechanism for encoding character code points into bit patterns may vary from entity to entity. All XML processors MUST accept the UTF-8 and UTF-16 encodings of Unicode

danielmarbach: That being said I think for this PR the discussion is not relevant. See my other comment

src/NServiceBus.Transport.SQS/TransportMessage.cs
jpalac: I have added a test for null.

src/NServiceBus.Transport.SQS/Extensions/MessageExtensions.cs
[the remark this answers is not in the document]
jpalac: I have made a change to use Array.Empty<byte>(). I got rid of the "empty message" placeholder concept and made the value null when there is no body. I have modified the tests to check that it works with the body being empty (ie "") as well as it being null.

src/NServiceBus.Transport.SQS/TransportMessage.cs
```diff
@@ -28,9 +28,9 @@ public TransportMessage(OutgoingMessage outgoingMessage, DispatchProperties prop
             }
 
 #if NETFRAMEWORK
-            Body = outgoingMessage.Body.Length != 0 ? Convert.ToBase64String(outgoingMessage.Body.ToArray()) : "empty message";
```
mikeminutillo, line 31: If someone has two endpoints today (Sender and Receiver) that are exchanging empty messages, this will break things if the sender is upgraded before the receiver. Is that likely (or even supported)?

src/NServiceBus.Transport.SQS.Tests/InputQueuePumpTests.cs
```diff
@@ -115,7 +115,7 @@ public async Task Expired_messages_are_deleted_without_processing()
                     {Headers.MessageId, messageId},
                     {TransportHeaders.TimeToBeReceived, ttbr.ToString()}
                 },
-                Body = "empty message"
```
mikeminutillo, line 118: Do we still need to support this type of "empty" message? If so, should we duplicate this test rather than modifying it? Or is the body irrelevant to this test?

src/NServiceBus.Transport.SQS.Tests/InputQueuePumpTests.cs
```diff
@@ -160,7 +160,7 @@ public async Task Processed_messages_are_deleted()
                 {
                     {Headers.MessageId, messageId}
                 },
-                Body = "empty message"
```
mikeminutillo, line 163: Same question as above

src/NServiceBus.Transport.SQS.Tests/InputQueuePumpTests.cs
jpalac: The body is irrelevant but I have used string.Empty and documented the reasoning here: https://github.com/Particular/Platform/pull/446

danielmarbach:
Rebased and added some minor tweaks. Would it be possible to add one or two acceptance tests?

jpalac:
Unfortunately found that [Amazon SQS requires the MessageBody to have at least 1 character](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html), so I have reverted back to using the "empty message" magic :-(

This will also make the [decision document](https://github.com/Particular/Platform/pull/446) irrelevant.

mikeminutillo (approved):

danielmarbach:
What if we did send `{}` instead? This would at least make some sense because the wrapper is json. We could byte encode this once and then always send that

jpalac:
> What if we did send `{}` instead? This would at least make some sense because the wrapper is json. We could byte encode this once and then always send that

Currently with the empty message, we don't base64 encode it, and then check for that string to also not decode it. If we get `{}`, couldn't that be a real value that is sent through by the user? In that case if they have encoding on, then we should be encoding it.

danielmarbach:
I think the same argument could be applied to "empty message". So in essence, it boils down to saying there is a representation of an empty message that is not base64 encoded. That is essentially what the transport did, and I think we are now discussing what the better representation of an empty message is.

jpalac:
Yes and we talked about it and thought that since we are already using "empty message" then keeping it this way is the path of least resistance in terms of wire compatibility.

PR #1892, opened 2023-02-02
Processing native message without base64 encoding on the body

jpalac:


src/NServiceBus.Transport.SQS/Extensions/MessageExtensions.cs
```diff
@@ -34,13 +38,40 @@ static class MessageExtensions
...
+            int contentLength = (int)s3GetResponse.ContentLength;
+            var buffer = arrayPool.Rent(contentLength);
+            using var memoryStream = new MemoryStream(buffer);
+            await s3GetResponse.ResponseStream.CopyToAsync(memoryStream, 81920, cancellationToken).ConfigureAwait(false);
+            return (buffer.AsMemory(0, contentLength), buffer);
+        }
+
+        static (ReadOnlyMemory<byte> MessageBody, byte[]? MessageBodyBuffer) ConvertBody(string body, ArrayPool<byte> arrayPool)
+        {
+            var encoding = Encoding.Unicode;
```
danielmarbach, line 50: I'm not entirely sure if we have to use UTF8 or if we can use `Unicode`

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html

links to https://www.w3.org/TR/REC-xml/#charsets

which says

> The mechanism for encoding character code points into bit patterns may vary from entity to entity. All XML processors MUST accept the UTF-8 and UTF-16 encodings of Unicode

mikeminutillo:
Closed in favor of #1905

PR #1906, opened 2023-02-06
Extract incoming message translation strategy

mikeminutillo:
Pulls the message translation strategy (from Amazon SQS message to NServiceBus transport message) into a separate class so that we can test it independently.


src/NServiceBus.Transport.SQS/IAmazonSqsMessageTranslationStrategy.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// Performs the SQS transport message extraction
+        /// </summary>
+        /// <param name="receivedMessage"></param>
+        /// <param name="messageId"></param>
+        /// <param name="headers"></param>
+        /// <param name="s3BodyKey"></param>
+        /// <param name="body"></param>
+        /// <returns></returns>
+        bool TryExtractMessage(Message receivedMessage, string messageId, out Dictionary<string, string> headers, out string s3BodyKey, out string body);
+    }
```
mauroservienti, line 22: this should go in a separate `IAmazonSqsIncomingMessageExtractor.cs` file

src/NServiceBus.Transport.SQS/IAmazonSqsMessageTranslationStrategy.cs
```diff
@@ -0,0 +1,62 @@
...
+
+    class DefaultAmazonSqsIncomingMessageExtractor : IAmazonSqsIncomingMessageExtractor
```
mauroservienti, line 24: What about making the `DefaultAmazonSqsIncomingMessageExtractor` public?
It would simplify users' life, they could handle only their special cases, and fallback to the default extractor if needed.

src/NServiceBus.Transport.SQS/IAmazonSqsMessageTranslationStrategy.cs
jpalac: Thanks - missed that after our initial run through - separated into 2 different files.

src/NServiceBus.Transport.SQS/IAmazonSqsMessageTranslationStrategy.cs
mikeminutillo: In theory a customer that wants to do this could already do so like this:

```csharp
sqs.IncomingMessageExtractor = new MyCustomIncomingMessageExtractor(sqs.IncomingMessageExtractor);
```

Is it worth adding this class to the public interface?

mauroservienti: Ah, good point. I missed that. No need to make it public.

src/NServiceBus.Transport.SQS/IAmazonSqsIncomingMessageExtractor.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace NServiceBus.Transport.SQS
+{
+    using System.Collections.Generic;
+    using Amazon.SQS.Model;
+
+    /// <summary>
+    /// Configures how the incoming SQS transport message is extracted
+    /// </summary>
+    public interface IAmazonSqsIncomingMessageExtractor
```
mauroservienti, line 9: As discussed today, we can rename this removing the reference to AmazonSQS. For example, `IIncomingMessageExtractor` or an even simpler `IMessageExtractor`.

jpalac: Changed to IMessageExtractor.

danielmarbach:
Closing in favour of https://github.com/Particular/NServiceBus.AmazonSQS/pull/1931

PR #1912, opened 2023-02-06
Set a wildcard policy on flaky tests

### LINKED ISSUE #1677: Fix flaky tests caused by SNS policy propagation delays
Some of the acceptance tests are written in such a way, that they can suffer from SNS access policy propagation delays. These are tests where more than one manual `Subscribe` operation is called:

- [Remove_multiple_events_when_setting_wildcard_policy](https://github.com/Particular/NServiceBus.AmazonSQS/blob/28c81ee2424d2a3c90f6e1aabe03cf32bbce08bb/src/CommandLineTests/CommandLineTests.cs#L360) - fails with error `System.InvalidOperationException : Process must exit before requested information can be determined`.
- Should_still_deliver - fails with `TestTimeout` exception

mauroservienti:
Fixes #1677

### DIFF
--- a/src/NServiceBus.Transport.SQS.AcceptanceTests/NativePubSub/When_concurrent_subscribe.cs
+++ b/src/NServiceBus.Transport.SQS.AcceptanceTests/NativePubSub/When_concurrent_subscribe.cs
@@ -12,7 +12,8 @@
         public async Task It_should_still_deliver()
         {
             var context = await Scenario.Define<Context>()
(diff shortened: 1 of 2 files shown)


danielmarbach (approved):

danielmarbach:
I think it would be great to bring those changes onto the supported release branches too.

danielmarbach:
Talked to Brandon and he asked if we could hold off merging the release branches until their work is done

danielmarbach:
Discussed this today in the call and we decided to move forward with this change even though we have seen the tests still occasionally failing. We concluded these changes make sense and will decrease the likelihood of the tests failing. We have to continue tracking the test failures and then in the future potentially re-evaluate things. Although we must say we couldn't come with a good solution to approach the eventual consistent nature of the policy propagation.

PR #1939, opened 2023-02-13
Untangle message preparation

danielmarbach:


src/NServiceBus.Transport.SQS/MessageDispatcher.cs
```diff
@@ -263,12 +263,9 @@ await sqsClient.SendMessageAsync(message.ToRequest(), cancellationToken)
...
         }
 
-        async Task<TMessage> PrepareMessage<TMessage>(IOutgoingTransportOperation transportOperation, Dictionary<string, Type> multicastEventsMessageIdsToType, TransportTransaction transportTransaction, CancellationToken cancellationToken)
-            where TMessage : PreparedMessage, new()
+        async Task<SqsPreparedMessage> PrepareMessage(UnicastTransportOperation transportOperation, Dictionary<string, Type> multicastEventsMessageIdsToType, TransportTransaction transportTransaction, CancellationToken cancellationToken)
         {
-            var unicastTransportOperation = transportOperation as UnicastTransportOperation;
-
-            if (!await hybridPubSubChecker.PublishUsingMessageDrivenPubSub(unicastTransportOperation, multicastEventsMessageIdsToType, topicCache, queueCache, snsClient).ConfigureAwait(false))
+            if (!await hybridPubSubChecker.PublishUsingMessageDrivenPubSub(transportOperation, multicastEventsMessageIdsToType, topicCache, queueCache, snsClient).ConfigureAwait(false))
```
jpalac, line 268: I know that this isn't a new method but I found the name `PublishUsingMessageDrivenPubSub` confusing in this context. This PrepareMessage is preparing any message for sending, not just publishing messages using message driven pub sub, so what we are really checking is more like:

```
if (await hybridPubSubChecker.ThisIsAPublishMessageNotUsingMessageDrivenPubSub)
{
   return null;
}
```

src/NServiceBus.Transport.SQS/MessageDispatcher.cs
```diff
@@ -341,24 +337,61 @@ async Task<TMessage> PrepareMessage<TMessage>(IOutgoingTransportOperation transp
...
+                {
+                    BucketName = s3.BucketName,
+                    InputStream = bodyStream,
+                    Key = key
+                };
+
+                s3.NullSafeEncryption.ModifyPutRequest(putObjectRequest);
+
+                await s3.S3Client.PutObjectAsync(putObjectRequest, cancellationToken).ConfigureAwait(false);
             }
```
jpalac, line 377: We should be able to clean up some of this repetition when we add in the `doNotWrapOutgoingMessages` changes. Certainly the pushing to s3 bit at least.

jpalac (commented):
I think it looks good - if we want to continue down this path then we should try and combine this with the `doNotWrapOutgoingMessages` changes sooner rather than later, as they are overlapping and I don't think a merge will be fun.

src/NServiceBus.Transport.SQS/MessageDispatcher.cs
danielmarbach: Agreed

danielmarbach:
Discussed with @jpalac and we will fix some of the outstanding comments in the next iteration on this code. Merging

jpalac (approved):

PR #1962, opened 2023-02-22
Add header for message body encoding type

jpalac:
This needs a test but we seem to be unable to send a unicode string from .Net - it converts unicode characters to utf-16 on the fly. Unless there's some low level way of doing this that we don't know about?


mauroservienti:
As far as I understand, we'll drop this one. Right?

jpalac:
Yes - perhaps just reference it on the AWS issue as a possible solution if anyone presents a decoding problem with native messages.

mauroservienti:
We decided not to introduce this for now. Closing.

PR #1963, opened 2023-02-22
Do not auto install

jpalac:
Queues only created if setupInfrastructure == true


mauroservienti:
The implementation for queues is good. We need to have the same for the `SubscriptionManager`. If they want to run the endpoint in "least privilege" mode, the endpoint won't have permission to create SNS topics, too.

What's tricky in the subscription manager is detecting the "it's the auto-subscribe" feature or the "it's the user code" explicitly trying to create a subscription. If installers are not enabled, both requests should be denied, but in the first case, we want to silently not create topics. In the latter, we want to throw.

danielmarbach:
What @mauroservienti said.

Unfortunately things are not super beautiful in the transport seam when it comes to discover that. Azure Service Bus does this trickery

https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/master/src/Transport/AzureServiceBusTransport.cs#L101-L118 and the subscription manager being null when `UsePublishSubscribe` is `false`
https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/a8a82a88035c09fe9b9a2bbbc49cd944f2f55eed/src/Transport/AzureServiceBusTransportInfrastructure.cs#L72-L75

but honestly I think this logic is flawed and should be revised. RabbitMQ seems to handle this better

https://github.com/Particular/NServiceBus.RabbitMQ/blob/master/src/NServiceBus.Transport.RabbitMQ/RabbitMQTransportInfrastructure.cs#L37-L55

Essentially it goe

mikeminutillo:
Azure Service Bus only sets up the original subscription to the shared topic. The Subscription Manager doesn't care if you have the flag set, it creates a rule on the subscription. The same is true of RabbitMQ. The Subscription Manager always creates the subscription.

mauroservienti:
If the AWS account doesn't have the following permissions:

- CreateTopic
- Subscribe
- Unsubscribe

The subscription manager will blow up at runtime. The same applies to S3 if the CreateBucket permission is missing.

danielmarbach:
@mikeminutillo yes but rule management would actually not require manage rights if we did it slightly differently. There is an issue in the repo about that

src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
```diff
@@ -226,12 +226,15 @@ public override async Task<TransportInfrastructure> Initialize(HostSettings host
...
+                var queueCreator = new QueueCreator(SqsClient, QueueCache, S3, maxTimeToLive, QueueDelayTime);
 
-            var createQueueTasks = sendingAddresses.Select(x => queueCreator.CreateQueueIfNecessary(x, false))
-                .Concat(infra.Receivers.Values.Select(x => queueCreator.CreateQueueIfNecessary(x.ReceiveAddress, true))).ToArray();
+                var createQueueTasks = sendingAddresses.Select(x => queueCreator.CreateQueueIfNecessary(x, false))
+                    .Concat(infra.Receivers.Values.Select(x => queueCreator.CreateQueueIfNecessary(x.ReceiveAddress, true))).ToArray();
 
-            await Task.WhenAll(createQueueTasks).ConfigureAwait(false);
+                await Task.WhenAll(createQueueTasks).ConfigureAwait(false);
+            }
```
mauroservienti, line 237: Nitpicking, feel free to ignore. Is it better to pass the `setupInfrastructure` flag in the `QueueCreator` ctor? It simplifies this code, and the `QueueCrerator` already exposed "if"-based methods, like `CreateQueueIfNecessary`.

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -45,6 +45,22 @@ class InputQueuePump : IMessageReceiver
...
         {
+            try
+            {
+                inputQueueUrl = await queueCache.GetQueueUrl(ReceiveAddress, cancellationToken)
+                .ConfigureAwait(false);
+            }
+            catch (QueueDoesNotExistException ex)
+            {
+                throw new QueueDoesNotExistException(
+                        $"Queue `{ReceiveAddress}` doesn't exist. Call endpointConfiguration.EnableInstallers() to create the queues at startup, or create them manually.",
```
mauroservienti, line 56: If we pass the `setupInfrastructure` flag to the pump, can we improve the exception message? E.g., when `setupInfrastructure` is true, we can throw an exception with a different message.

mauroservienti (approved):

src/NServiceBus.Transport.SQS/InputQueuePump.cs
jpalac: I have passed the flag to the pump and adjusted the message accordingly.

src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
jpalac: I think it's clearer this way and in line with how the other transports do it. Also, it saves us having to create an instance of QueueCreator.

src/NServiceBus.Transport.SQS.AcceptanceTests/When_trying_to_subscribe_when_topic_does_not_exist.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        public class MyMessage : ICommand
+        {
+        }
+
+        public class MyEvent : IEvent
+        {
+        }
+
+        public class DefaultServerWithNoInstallers : ServerWithInstallersDisabled
```
danielmarbach, line 84: Let's move this out into a seperate file

src/NServiceBus.Transport.SQS.AcceptanceTests/When_trying_to_subscribe_when_topic_does_not_exist.cs
```diff
@@ -0,0 +1,123 @@
...
+            public override Task<EndpointConfiguration> GetConfiguration(RunDescriptor runDescriptor, EndpointCustomizationConfiguration endpointConfiguration, Func<EndpointConfiguration, Task> configurationBuilderCustomization) =>
+                base.GetConfiguration(runDescriptor, endpointConfiguration, async configuration =>
+                {
+                    await configuration.DefinePersistence(PersistenceConfiguration, runDescriptor, endpointConfiguration);
+
+                    await configurationBuilderCustomization(configuration);
+                });
+        }
+
+        public class ServerWithInstallersDisabled : IEndpointSetupTemplate
```
danielmarbach, line 97: Let's move this out into a seperate file

src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
```diff
@@ -226,12 +226,15 @@ public override async Task<TransportInfrastructure> Initialize(HostSettings host
...
             var infra = new SqsTransportInfrastructure(this, hostSettings, receivers, SqsClient, SnsClient, QueueCache, topicCache, S3, Policies, QueueDelayTime, topicNamePrefix, EnableV1CompatibilityMode, DoNotWrapOutgoingMessages);
 
-            var queueCreator = new QueueCreator(SqsClient, QueueCache, S3, maxTimeToLive, QueueDelayTime);
+            if (hostSettings.SetupInfrastructure)
+            {
+                var queueCreator = new QueueCreator(SqsClient, QueueCache, S3, maxTimeToLive, QueueDelayTime);
 
-            var createQueueTasks = sendingAddresses.Select(x => queueCreator.CreateQueueIfNecessary(x, false))
-                .Concat(infra.Receivers.Values.Select(x => queueCreator.CreateQueueIfNecessary(x.ReceiveAddress, true))).ToArray();
+                var createQueueTasks = sendingAddresses.Select(x => queueCreator.CreateQueueIfNecessary(x, false))
```
danielmarbach, line 233: I think `CreateQueueIfNecessary` accepts a cancellation token. Should we pass it?

src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
```diff
@@ -226,12 +226,15 @@ public override async Task<TransportInfrastructure> Initialize(HostSettings host
...
+                    .Concat(infra.Receivers.Values.Select(x => queueCreator.CreateQueueIfNecessary(x.ReceiveAddress, true))).ToArray();
```
danielmarbach, line 234: I think `CreateQueueIfNecessary` accepts a cancellation token. Should we pass it?

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -41,10 +42,32 @@ class InputQueuePump : IMessageReceiver
...
         public async Task Initialize(PushRuntimeSettings limitations, OnMessage onMessage, OnError onError, CancellationToken cancellationToken = default)
         {
+            try
+            {
+                inputQueueUrl = await queueCache.GetQueueUrl(ReceiveAddress, cancellationToken)
+                .ConfigureAwait(false);
+            }
+            catch (QueueDoesNotExistException ex)
+            {
+                var msg = $"Queue `{ReceiveAddress}` doesn't exist. Call endpointConfiguration.EnableInstallers() to create the queues at startup, or create them manually.";
```
danielmarbach, line 57: Mauro: We can probably avoid creating the string based on the setup infrastructure flag

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -41,10 +42,32 @@ class InputQueuePump : IMessageReceiver
...
             this.coreSettings = coreSettings;
+            this.setupInfrastructure = setupInfrastructure;
         }
 
         public async Task Initialize(PushRuntimeSettings limitations, OnMessage onMessage, OnError onError, CancellationToken cancellationToken = default)
         {
+            try
+            {
+                inputQueueUrl = await queueCache.GetQueueUrl(ReceiveAddress, cancellationToken)
+                .ConfigureAwait(false);
```
danielmarbach, line 53: Weird indendation

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -590,6 +613,9 @@ async Task MovePoisonMessageToErrorQueue(Message message, CancellationToken mess
         readonly S3Settings s3Settings;
         readonly Action<string, Exception, CancellationToken> criticalErrorAction;
         readonly IReadOnlySettings coreSettings;
+
```
danielmarbach, line 616: Nit: :fire:

src/NServiceBus.Transport.SQS.AcceptanceTests/When_trying_to_subscribe_when_topic_does_not_exist.cs
```diff
@@ -0,0 +1,123 @@
...
+    using NServiceBus.AcceptanceTesting.Support;
+    using NServiceBus.Features;
+    using NUnit.Framework;
+
+    public class When_trying_to_subscribe_when_topic_does_not_exist : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_log_exception_with_topic_name()
+        {
+            var messageId = Guid.NewGuid();
```
danielmarbach, line 18: The message id is not relevant for this test right? Can we kill it?

src/NServiceBus.Transport.SQS.AcceptanceTests/When_trying_to_subscribe_when_topic_does_not_exist.cs
```diff
@@ -0,0 +1,123 @@
...
+
+            await Scenario.Define<Context>(c =>
+            {
+                c.MessageId = messageId;
+                c.EnableInstallers = true;
+                c.DisableAutoSubscribe = true;
+            })
+            .WithEndpoint<Endpoint>()
+            .Done(context => true)
```
danielmarbach, line 27: The done condition looks weird. Should we switch to context.EndpointsStarted?

src/NServiceBus.Transport.SQS.AcceptanceTests/When_trying_to_subscribe_when_topic_does_not_exist.cs
```diff
@@ -0,0 +1,123 @@
...
+            .Run();
+
+            var context = await Scenario.Define<Context>(c =>
+            {
+                c.MessageId = messageId;
+                c.EnableInstallers = false;
+                c.DisableAutoSubscribe = false;
+            })
+            .WithEndpoint<Endpoint>()
+            .Done(context => true)
```
danielmarbach, line 37: The done condition looks weird. Should we switch to context.EndpointsStarted?

src/NServiceBus.Transport.SQS.AcceptanceTests/When_trying_to_subscribe_when_topic_does_not_exist.cs
```diff
@@ -0,0 +1,123 @@
...
+            .Run();
+
+            Assert.IsTrue(context.Logs.Any(a => a.Message.Contains("not found") && a.Message.Contains("When_trying_to_subscribe_when_topic_does_not_exist-MyEvent")));
```
danielmarbach, line 40: How about we assert the log level too?

Could we use a more expressive collection assert with a predicate? Something like

Assert.That(context.Logs, Has.One.Matches<ScenarioContext.LogItem>(x => x.Level ==  && ));

or whatever the NUnit thing is. Because that would make sure we get a more meaning message when the assertion fails.

src/NServiceBus.Transport.SQS.AcceptanceTests/When_trying_to_subscribe_when_topic_does_not_exist.cs
```diff
@@ -0,0 +1,123 @@
...
+            public bool EnableInstallers { get; set; }
+            public bool DisableAutoSubscribe { get; set; }
+        }
+
+        public class Endpoint : EndpointConfigurationBuilder
+        {
+            public Endpoint() =>
+                EndpointSetup<DefaultServerWithNoInstallers>((e, c) =>
+                {
+                    var ourContext = c.ScenarioContext as Context;
```
danielmarbach, line 55: testContext or scenarioContext instead of ourContext?

src/NServiceBus.Transport.SQS.AcceptanceTests/When_trying_to_subscribe_when_topic_does_not_exist.cs
```diff
@@ -0,0 +1,123 @@
...
+                    if (ourContext.EnableInstallers)
+                    {
+                        e.EnableInstallers();
+                    }
+                    if (ourContext.DisableAutoSubscribe)
+                    {
+                        e.DisableFeature<AutoSubscribe>();
+                    }
```
danielmarbach, line 63: Isn't this plus the two scenario define calls an indication we should split into two Shoulds or potentially even whens?

src/NServiceBus.Transport.SQS.AcceptanceTests/When_trying_to_receive_message_from_nonexisting_queue.cs
```diff
@@ -0,0 +1,85 @@
...
+            var messageId = Guid.NewGuid();
+
+            var exception = Assert.ThrowsAsync<QueueDoesNotExistException>(async () =>
+            {
+                await Scenario.Define<Context>(c =>
+                    {
+                        c.MessageId = messageId;
+                    })
+                    .WithEndpoint<Endpoint>()
+                    .Done(context => true)
```
danielmarbach, line 26: Some question about the done as below

src/NServiceBus.Transport.SQS.AcceptanceTests/When_trying_to_receive_message_from_nonexisting_queue.cs
```diff
@@ -0,0 +1,85 @@
...
+    using NServiceBus.AcceptanceTesting.Customization;
+    using NServiceBus.AcceptanceTesting.Support;
+    using NUnit.Framework;
+
+    public class When_trying_to_receive_message_from_nonexisting_queue : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public void Should_throw_exception_with_queue_name()
+        {
+            var messageId = Guid.NewGuid();
```
danielmarbach, line 17: Same question about the message id as below

src/NServiceBus.Transport.SQS.AcceptanceTests/When_trying_to_subscribe_when_topic_does_not_exist.cs
```diff
@@ -0,0 +1,123 @@
...
+    using System.Linq;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NServiceBus.AcceptanceTesting.Customization;
+    using NServiceBus.AcceptanceTesting.Support;
+    using NServiceBus.Features;
+    using NUnit.Framework;
+
+    public class When_trying_to_subscribe_when_topic_does_not_exist : NServiceBusAcceptanceTest
```
danielmarbach, line 13: Should this test be moved into the NativePubsub folder?

src/NServiceBus.Transport.SQS.AcceptanceTests/When_trying_to_subscribe_when_topic_does_not_exist.cs
jpalac: What we need here is for the topic to not be created - we want to test the error message when someone is subscribing to a topic that doesn't exist.
 In order to test that, we need the installers off. However, when that happens, an exception will be thrown earlier on in the pipeline during the `InputQueuePump.Initialize` call, before we can test our subscription.
So what we are doing is creating the first endpoint with installers but without auto subscribe so that `SubscribeAll `is not called and hence the topic is not created.
We then create a second endpoint that has the installers disabled but subscibers on. So, it will not throw the exception during the `InputQueuePump.Initialize` call since the queue was created by the previous endpoint, but it will at the `SubscribeAll` point since the topic does not exist and installers are off.

From what I can see we can't call a `.WithEndpoint` based on a `When` - ie don't start the second endpoint until the first one is started. But I could be wrong!

src/NServiceBus.Transport.SQS/SubscriptionManager.cs
```diff
@@ -131,32 +132,39 @@ async Task CreateTopicAndSubscribe(string topicName, string queueUrl, Concurrent
...
-
-        async Task CreateTopicAndSubscribe(Type eventType, string queueUrl, ConcurrentBag<PolicyStatement> policyStatementsToBeSettled, CancellationToken cancellationToken)
-        {
-            var topicName = topicCache.GetTopicName(eventType);
-
-            if (Logger.IsDebugEnabled)
+            else
             {
-                Logger.Debug($"Getting or creating topic '{topicName}' for queue '{queueName}");
+                var getTopicResponse = await snsClient.FindTopicAsync(topicName).ConfigureAwait(false);
```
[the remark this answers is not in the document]
jpalac: We only have the topic name at this point. The topic cache is using the type everywhere as the key.

src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
jpalac: I have passed it in

src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
jpalac: I have passed it in

src/NServiceBus.Transport.SQS.AcceptanceTests/When_trying_to_subscribe_when_topic_does_not_exist.cs
jpalac: I have moved both of those into separate files and the relevant tests into a new folder 'Installers'

src/NServiceBus.Transport.SQS.AcceptanceTests/When_trying_to_subscribe_when_topic_does_not_exist.cs
jpalac: renamed

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -41,10 +42,31 @@ class InputQueuePump : IMessageReceiver
...
+            {
+                inputQueueUrl = await queueCache.GetQueueUrl(ReceiveAddress, cancellationToken).ConfigureAwait(false);
+            }
+            catch (QueueDoesNotExistException ex)
+            {
+                var msg = $"Queue `{ReceiveAddress}` doesn't exist. Call endpointConfiguration.EnableInstallers() to create the queues at startup, or create them manually.";
+                if (setupInfrastructure)
+                {
+                    msg = $"Queue `{ReceiveAddress}` doesn't exist. Ensure this process has the required permissions to create queues on Amazon SQS.";
+                }
```
mauroservienti, line 60: We can do something like this:

```suggestion
                var msg = setupInfrastructure
                    ? $"Queue `{ReceiveAddress}` doesn't exist. Ensure this process has the required permissions to create queues on Amazon SQS."
                    :  $"Queue `{ReceiveAddress}` doesn't exist. Call endpointConfiguration.EnableInstallers() to create the queues at startup, or create them manually.";
```

danielmarbach (approved):
Looks with the outstanding comment from Mauro

mikeminutillo (approved):

PR #1968, opened 2023-02-23
[Release 5.7]  Added no message wrapper option

mikeminutillo:
Backport #1943 to release 5.7


src/NServiceBus.Transport.SQS.AcceptanceTests/Publishing/When_publishing_wrapped_event.cs
```diff
@@ -22,7 +22,7 @@ public async Task Should_receive_event(byte[] payload)
                 .WithEndpoint<Publisher>(b => b.When(c => c.EndpointsStarted, session => session.Publish(new MyEventWithPayload() { Payload = payload })))
                 .WithEndpoint<Subscriber>(b => { })
                 .Done(c => c.GotTheEvent)
-                .Run(TimeSpan.FromSeconds(30));
+                .Run(TimeSpan.FromSeconds(45));
```
mauroservienti, line 25: What was the reason for making the timeout longer?

src/NServiceBus.Transport.SQS/Configure/SqsTransportInfrastructure.cs
```diff
@@ -98,7 +98,8 @@ QueueCreator CreateQueueCreator()
 
         MessageDispatcher CreateMessageDispatcher()
         {
-            return new MessageDispatcher(configuration, s3Client, sqsClient, snsClient, queueCache, topicCache);
+            var doNotWrapOutgoingMessages = settings.GetOrDefault<bool>(SettingsKeys.DoNotWrapOutgoingMessages);
+            return new MessageDispatcher(configuration, s3Client, sqsClient, snsClient, queueCache, topicCache, !doNotWrapOutgoingMessages);
```
mauroservienti, line 102: The not here makes the code harder to understand. What was the reason for using it?

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -286,7 +306,7 @@ async Task<bool> InnerProcessMessage(Dictionary<string, string> headers, string
                     var messageContext = new MessageContext(
                         nativeMessageId,
                         new Dictionary<string, string>(headers),
-                        body,
+                        body.ToArray(),
```
danielmarbach, line 309: This should not be necessary and actually invalidates the whole buffer pooling. The pool returns a byte array and the code that targets v8 just treats it `AsMemory`. In order to still benefit from the pooling approach, can simplify the code a bit for v7 to only use the `byte[]` reference.

src/NServiceBus.Transport.SQS.AcceptanceTests/Publishing/When_publishing_wrapped_event.cs
jpalac: It kept getting a timeout error when running the tests on the build server.

src/NServiceBus.Transport.SQS/Configure/SqsTransportInfrastructure.cs
jpalac: This is the same as on master. Since the default behaviour is to wrap the messages, the setting introduced is doNotWrapOutgoingMessage. In code we want to deal with the opposite of that (ie wrapOutgoingMessages) so when we pass it into the dispatcher we switch it over to that.

src/NServiceBus.Transport.SQS/InputQueuePump.cs
jpalac: I have changed it to use bye[] in general - as it was doing originally.

src/NServiceBus.Transport.SQS/Extensions/TransportMessageExtensions.cs
```diff
@@ -43,33 +41,27 @@ static class TransportMessageExtensions
...
-                var buffer = arrayPool.Rent(length);
-                var writtenBytes = encoding.GetBytes(body, 0, body.Length, buffer, 0);
-                return Tuple.Create((ReadOnlyMemory<byte>)buffer.AsMemory(0, writtenBytes), buffer);
+                return encoding.GetBytes(body);
             }
         }
 
-        static readonly Tuple<ReadOnlyMemory<byte>, byte[]>
-            EmptyMessage = Tuple.Create(new ReadOnlyMemory<byte>(), default(byte[]));
+        static readonly byte[] EmptyMessage = new byte[0];
```
danielmarbach, line 65: Array.Empty would work too. If you want to keep the static field due to expressiveness you could assign Array.Empty instead of new byte[0]

src/NServiceBus.Transport.SQS/Configure/SqsTransportInfrastructure.cs
danielmarbach: :exploding_head: But yeah it is also like that on master. Discussed with @mauroservienti and resolved it.

src/NServiceBus.Transport.SQS/InputQueuePump.cs
danielmarbach: Awesome

PR #1969, opened 2023-02-24
Ensure 64 decoding doesn't happen on native message

jpalac:
Only try to 64decode if not a native message


mikeminutillo (approved):

danielmarbach (commented):
:+1:

PR #1973, opened 2023-02-26
Set HttpClientCacheSize to 1 instead of relying on the default value

danielmarbach:
Some more context in https://github.com/aws/aws-sdk-net/issues/1929

I have also submitted [a PR to fix this in the SDK](https://github.com/aws/aws-sdk-net/pull/2549).


danielmarbach:
Will address

- Log Warning if the external client has different settings

in a dedicated PR

PR #1976, opened 2023-02-27
Update dotnet framework to 462

jpalac:


danielmarbach (changes requested):
I think we need to drop netcoreapp3.1 and change the ci yaml
I would also suggest to target directly the release branch since this change us independent of the other

jpalac:
Closing in favour of https://github.com/Particular/NServiceBus.AmazonSQS/pull/1977

PR #1984, opened 2023-02-28
Do not dispose externally managed clients

### LINKED ISSUE #1974: Externally managed clients are disposed
The transport disposes all clients regardless of whether they are created by the transport or passed in externally. The transport is free to dispose clients it manages but should not be disposing clients it doesn't own since the clients of the transport might decide to continue using the clients they passed to the transport even after the transport was stopped.

mauroservienti:
Fix #1974 

- [x] Rebase
- [x] Tests
- [x] Documentation Added to this [PR](https://github.com/Particular/docs.particular.net/pull/6001)

### DIFF
--- a/src/NServiceBus.Transport.SQS/Configure/SqsTransportInfrastructure.cs
+++ b/src/NServiceBus.Transport.SQS/Configure/SqsTransportInfrastructure.cs
@@ -14,12 +14,13 @@
     {
         public SqsTransportInfrastructure(SqsTransport transportDefinition, HostSettings hostSettings, ReceiveSettings[] receiverSettings, IAmazonSQS sqsClient,
             IAmazonSimpleNotificationService snsClient, QueueCache queueCache, TopicCache topicCache, S3Settings s3Settings, PolicySettings policySettings, int queueDelayTimeSeconds, string topicNamePrefix, bool v1Compatibility, bool doNotWrapOutgoingMessages,
-            bool shouldDisposeSqsAndSnsClients)
+            bool shouldDisposeSqsClient, bool shouldDisposeSnsClient)
         {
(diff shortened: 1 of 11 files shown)


jpalac (approved):

mikeminutillo:
Because of the old transport configuration API it is possible for these to be set after the constructor is called. In theory a user can override just one of the client factories so we need to split the flag for SQS/SNS.

src/NServiceBus.Transport.SQS/Configure/SqsTransportInfrastructure.cs
```diff
@@ -56,8 +71,11 @@ public override Task Shutdown(CancellationToken cancellationToken = default)
         readonly SqsTransport transportDefinition;
         readonly IAmazonSQS sqsClient;
         readonly IAmazonSimpleNotificationService snsClient;
+        readonly bool shouldDisposeSqsClient;
+        readonly bool shouldDisposeSnsClient;
         readonly IAmazonS3 s3Client;
         readonly IReadOnlySettings coreSettings;
         readonly bool setupInfrastructure;
+        readonly bool shouldDisposeS3Client;
```
mauroservienti, line 79:
```suggestion
        readonly IAmazonS3 s3Client;
        readonly IReadOnlySettings coreSettings;
        readonly bool setupInfrastructure;
        readonly bool shouldDisposeSqsClient;
        readonly bool shouldDisposeSnsClient;
        readonly bool shouldDisposeS3Client;
```

mikeminutillo:
@mauroservienti at the moment it is allowed to pass in `null` for any of the clients and we don't guard against that anywhere. Is there a good reason for that or should we be guarding against `null` values?

danielmarbach (approved):

PR #2003, opened 2023-03-02
Support for batch publishing

danielmarbach:
This pull request leverages previously done refactoring to introduce batching for SNS publishes.

### DIFF
--- a/src/NServiceBus.Transport.SQS/SnsPreparedMessageBatcher.cs
+++ b/src/NServiceBus.Transport.SQS/SnsPreparedMessageBatcher.cs
@@ -11,7 +11,7 @@ namespace NServiceBus.Transport.SQS
         public static IReadOnlyList<SnsBatchEntry> Batch(IEnumerable<SnsPreparedMessage> preparedMessages)
         {
             var allBatches = new List<SnsBatchEntry>();
-            var currentDestinationBatches = new Dictionary<string, SnsPreparedMessage>();
+            var currentDestinationBatches = new Dictionary<string, SnsPreparedMessage>(TransportConstraints.MaximumItemsInBatch);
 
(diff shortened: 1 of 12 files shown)


src/NServiceBus.Transport.SQS/SnsPreparedMessageBatcher.cs
```diff
@@ -0,0 +1,64 @@
...
+                        currentDestinationBatches.Clear();
+                        payloadSize = size;
+                    }
+
+                    // we don't have to recheck payload size here because the support layer checks that a request can always fit 256 KB size limit
+                    // we can't take MessageId because batch request ID can only contain alphanumeric characters, hyphen and underscores, message id could be overloaded
+                    currentDestinationBatches.Add(Guid.NewGuid().ToString(), message);
+
+                    var currentCount = currentDestinationBatches.Count;
+                    if (currentCount != 0 && currentCount % TransportConstraints.MaximumItemsInBatch == 0)
```
jpalac, line 46: Just a curiosity question - how come you use the reminder operator instead of checking if currentCount == maxItems? Is it more efficient?

jpalac (approved):
Looks good

src/NServiceBus.Transport.SQS/SnsPreparedMessageBatcher.cs
danielmarbach: Good question. Actually I don't know why I used this style. I'll simplify it

danielmarbach:
@mauroservienti FYI I think we discussed this previously and mentioned that while batching increases the complexity of the SNS path we are already familiar with the batching APIs for SQS and even when publishing a single event the overhead of the batching code is minimal (similar for all the cases when users mostly send one message). 

I also have seen quite a few cases when events got abused and used in cases where a command should have been used. In those data shipping or integration scenarios many events are published.

PR #2123, opened 2023-04-20
Auto approve AWS dependencies and separate them to reduce the likelihood of merge conflicts

mauroservienti:
AWS never broke compatibility and never caused issues in updates so far. It’s safe enough to auto merge dependencies


DavidBoike (approved):

danielmarbach:
Not that it really matters but I was surprised we used an item group per dependency instead of one dedicated for all the auto merged ones with a label on that group that explains how we treat them

DavidBoike:
That is to keep them more than 3 lines apart so that merging one doesn’t require a rebase and a rerun of CI.

PR #2212, opened 2023-06-27
Reduce logging output when stopping receivers

timbussmann:
Stopping receivers usually ends up in an OperationCancelledException from the passed cancellation token. This is fully expected behavior. However, the current implementation generates enormous amounts of log output due to the very long stack traces being logged. These stack traces aren't really relevant as they are expected and don't provide any useful information but make the CI output very noisy as the few (38) transport tests cause about 4k lines of log entries.

### DIFF
--- a/src/NServiceBus.Transport.SQS/DelayedMessagesPump.cs
+++ b/src/NServiceBus.Transport.SQS/DelayedMessagesPump.cs
@@ -126,7 +126,7 @@ namespace NServiceBus.Transport.SQS
                 catch (Exception ex) when (ex.IsCausedBy(cancellationToken))
                 {
                     // private token, pump is being stopped, log the exception in case the stack trace is every required for debugging
-                    Logger.Debug("Operation canceled while stopping delayed message pump.", ex);
+                    Logger.Debug("Operation canceled while stopping delayed message pump.");
                     break;
                 }
                 catch (Exception ex)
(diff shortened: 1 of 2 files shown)


danielmarbach (approved):

kentdr:
I agree with your reasoning for how you have it written.  But, is it possible if a customer's system is not written to handle shutdowns well, that the stack trace is useful to find if a zombie/ghost record was created or some weird modified state issues due to an unexpected shutdown?

danielmarbach:
@kentdr what would they see in that stack trace given the catch is conditional, deep within our code and only executed when a specific token triggered?

```csharp
when (ex.IsCausedBy(messagePumpCancellationToken))
```

kentdr:
It would just show where the token was triggered and if it was before or after any state mutations.  My suggestion is contrived, but I felt it was worth mentioning because the stack trace is usually one of the first things we ask for when we are debugging.

timbussmann:
> But, is it possible if a customer's system is not written to handle shutdowns well, that the stack trace is useful to debug zombie/ghost records or some weird modified state issues due to an unexpected shutdown?

not sure about the "customer's system is not written to handle shutdowns well" - there is nothing to handle from a user perspective since we swallow the exception anyway. The stack trace is currently logged as part of a regular shutdown (depending on where there code currently is sitting) so I think it has even potentially negative effects by confusing users and making them think that something is wrong when it isn't.

kentdr (approved):

PR #2288, opened 2023-09-20
Fix native integration when using localstack instead of real AWS cloud

maddiehermann:
This PR fixes a bug that prevents native integration from working when using [localstack](https://github.com/localstack/localstack), a local AWS cloud stack implementation.

There is an undocumented feature (or perhaps an inconsistency) in the `MessageAttributeNames` parameter for `receiveMessage` in AWS.

The [official documentation](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SQS.html#receiveMessage-property) states that:

> you can return all of the attributes by specifying All or .* in your request

The undocumented feature is that AWS actually _also_ allows `*` to return all `MessageAttributes`.

`NServiceBus.AmazonSQS` [makes use](https://github.com/Particular/NServiceBus.AmazonSQS/blob/44acb865f0b34fd528808b65f61d3bea269f91e5/src/NServiceBus.Transport.SQS/InputQueuePump.cs#L129C15-L129C15) of this undocumented feature:

image

This works perfectly fine when working with the actual AWS cloud. But localstack seems to have written their implementation based on what the official AWS docs actually say, i.e. only `All` and `.*` (_not_ `*`*) will return all `MessageAttributes`. This means that even when you specify `MessageTypeFullName` to signal that it's a native message, the `InputQueuePump` doesn't see _any_ `MessageAttributes`, so it doesn't try to process it as a native message.

I think it makes sense to fix this in `NServiceBus.AmazonSQS` so that it doesn't rely on an undocumented feature. 

### Real AWS returns `MessageAttributes` for `*`, `.*` and `All`

image
image
image

### Localstack doesn't return `MessageAttributes` for `*`, but does for `.*` and `All`

image
image
image

### DIFF
--- a/src/NServiceBus.Transport.SQS/InputQueuePump.cs
+++ b/src/NServiceBus.Transport.SQS/InputQueuePump.cs
@@ -126,7 +126,7 @@ namespace NServiceBus.Transport.SQS
                 QueueUrl = inputQueueUrl,
                 WaitTimeSeconds = 20,
                 AttributeNames = new List<string> { "SentTimestamp" },
-                MessageAttributeNames = new List<string> { "*" }
+                MessageAttributeNames = new List<string> { ".*" }
             };
 
             if (coreSettings != null && coreSettings.TryGet<int>(SettingsKeys.MessageVisibilityTimeout, out var visibilityTimeout))


mauroservienti:
Thanks for raising this, @maddiehermann. Due to security policies, contributors' PRs cannot trigger CI. I raised https://github.com/Particular/NServiceBus.AmazonSQS/pull/2289 to bring this forward and run it through CI.

danielmarbach:
@maddiehermann on which version of the transport are you running? This adjustment doesn't seem to fall under our regular back porting policy so we might only patch the latest supported version. Would that work for you?

danielmarbach:
Closed in favor of https://github.com/Particular/NServiceBus.AmazonSQS/pull/2289

Could you reach out to mauro dot servienti at particular dot net? We would love to send you a gift of our appreciation for your first contribution.

maddiehermann:
Thanks @mauroservienti 

@danielmarbach We're currently on version 5.6.1, but that's ok if it's only patched in the latest version - we're considering upgrading in the near future so that we can use native integration without Base64 encoding the body, and this wildcard issue only affects our local testing. Plus I can work around it by adding the transport as a project reference and editing the wildcard code myself.

As for the gift, I've actually made 2 other small contributions 3 years ago, so it's not my first contribution 😊 But I wouldn't say no if you still wanted to show appreciation? 😀

mauroservienti:
> But I wouldn't say no if you still wanted to show appreciation? 😀

We'd love to do so, @maddiehermann. Please contact me at mauro.servienti [at] particular.net

mauroservienti:
@maddiehermann, I released [5.7.3](https://github.com/Particular/NServiceBus.AmazonSQS/releases/tag/5.7.3) and [6.1.2](https://github.com/Particular/NServiceBus.AmazonSQS/releases/tag/6.1.2), right now. They should be soon available on NuGet (time for indexing to catch up). Thanks for your contribution. Please, do not forget about 😜 👇 

> Please contact me at mauro.servienti [at] particular.net

Issue #2314, opened 2023-10-16
The QueueNameGenerator API might be misleading


SzymonPobiega:
#### Is your improvement related to a problem? Please describe.

The `QueueNameGenerator` API accepts a function of two arguments, the queue prefix and the queue name, and returns the final queue name with added prefix. The transport code assumes this function is idempotent i.e. that it can detect that a given queue name already has prefix and in that case skip applying it again. This fact is not documented which may lead to problems when trying to use the API.

On the flip side, the default queue name generator uses simple `Substring` to check if the prefix has already been applied leading to problems if the configured prefix is a substring of an endpoint name e.g. when prefix is `Prod` (denoting environment) and the endpoint name is `ProductionProcessControl`. In that scenario the correct queue name is `ProdProductionProcessControl` but the actual result is going to be `ProductionProcessControl`.

#### Describe the suggested solution

- Document the limitations of the API
- Introduce a two-part addressing scheme similar to MSMQ or SQL Server transports where the transport address includes both prefix and the queue name (delimited)

### Additional Context

Similar issue _might_ be affecting the `TopicNameGenerator` API (but it has not been verified)

mauroservienti:
This has been improved in https://github.com/Particular/NServiceBus.AmazonSQS/pull/2342. The transport at startup exercises the QueueNameGenerator API twice with the same values and validates the result is the same; otherwise, it throws.

PR #2321, opened 2023-10-20
Detect non-idempotent custom queue name generator

mikeminutillo:
- Partially addresses #2314

It is possible that the queue name generator may be applied multiple times. The default queue name generator is idempotent (see test in the PR) but a custom one may not be. This can lead to incorrect results when a custom queue name generator has been applied.

This PR adds a validation check when the custom queue name generator is set to ensure that applying it twice is safe.

### DIFF
--- a/src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
+++ b/src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
@@ -272,7 +272,7 @@
             var twice = generator(once, prefix);
             if (once != twice)
             {
-                throw new Exception($"Queue name generator is not idempotent. Result of applying twice {twice}");
+                throw new Exception($"The queue name generator function needs to return the same result when it is applied multiple times (idempotent). Result of applying once is {once} and twice -- {twice}.");
             }
         }
 
--- a/src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
+++ b/src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
@@ -67,7 +67,21 @@
             set
             {
                 Guard.ThrowIfNull(value);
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
```diff
@@ -67,10 +67,24 @@ internal set
             set
             {
                 Guard.ThrowIfNull(value);
+                AssertQueueNameGeneratorIdempotent(value);
```
mauroservienti, line 70: I'd move this to the transport runtime part. If the user-provided queue name generator depends on stuff available only when the endpoint is running (e.g., something in the IoC container), this will always fail even if the implementation is idempotent.

src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
```diff
@@ -275,6 +263,19 @@ public override async Task<TransportInfrastructure> Initialize(HostSettings host
...
+        static void AssertQueueNameGeneratorIdempotent(Func<string, string, string> generator)
+        {
+            const string prefix = "Prefix";
+            const string destination = "Destination";
+
+            var once = generator(destination, prefix);
+            var twice = generator(once, prefix);
+            if (once != twice)
+            {
+                throw new Exception($"Queue name generator is not idempotent. Result of applying twice {twice}");
```
SzymonPobiega, line 275:
```suggestion
                throw new Exception($"The queue name generator function needs to return the same result when it is applied multiple times (idempotent). Result of applying once is {once} and twice -- {twice}.");
```

mauroservienti (approved):

SzymonPobiega (approved):

PR #2331, opened 2023-11-02
Isolate transport message extraction

mikeminutillo:
Isolate and test the transport message extraction algorithm


mauroservienti:
@mikeminutillo, you can rebase on `receive-native-messages-with-no-nsb-attributes`; it should solve the poison message failing tests.

PR #2518, opened 2024-07-11
Add internal property to disable delay delivery

johnsimons:
This workaround allows ServiceControl to disable delay delivery for non-main message pumps.

This is the same workaround already in use for [SqlServer in ServiceControl](https://github.com/Particular/ServiceControl/blob/56f29e8fe68a51acb770153b7d140473d58c20fc/src/ServiceControl.Transports.SqlServer/SqlServerTransportCustomization.cs#L76-L77).

Property in SQLServer Transport: https://github.com/Particular/NServiceBus.SqlServer/blob/master/src/NServiceBus.Transport.SqlServer/SqlServerTransport.cs#L159
Usage of the property to determine delayed delivery on/off: https://github.com/Particular/NServiceBus.SqlServer/blob/master/src/NServiceBus.Transport.SqlServer/SqlServerTransportInfrastructure.cs#L161

### DIFF
--- a/src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
+++ b/src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
@@ -58,6 +58,11 @@
         public string QueueNamePrefix { get; set; }
 
         /// <summary>
+        /// Disable native delayed delivery infrastructure
+        /// </summary>
+        internal bool DisableDelayedDelivery { get; set; } = false;
+
+        /// <summary>
         /// Specifies a lambda function that allows to take control of the queue name generation logic.
         /// This is useful to overcome any limitations imposed by SQS.
         /// </summary>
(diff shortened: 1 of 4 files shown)


jpalac (approved):

johnsimons:
@mauroservienti @ramonsmits, this is the fix we are proposing to get ServiceControl back working.
I know it is just a workaround, but a real fix will need more thinking, whether we should allow for consumers to turn on/off delays for the transport.
This is the same "workaround" that already exists in [SqlServer transport](https://github.com/Particular/ServiceControl/blob/56f29e8fe68a51acb770153b7d140473d58c20fc/src/ServiceControl.Transports.SqlServer/SqlServerTransportCustomization.cs#L76-L77), so at least we are consistent with workarounds :wink:

So the plan, once we release this, is to update SC and toggle the flag. 
We have tested this against SC, and it works.

src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
```diff
@@ -57,6 +57,11 @@ internal set
         /// </summary>
         public string QueueNamePrefix { get; set; }
 
+        /// <summary>
+        /// Disable native delayed delivery infrastructure
```
mauroservienti, line 61: Maybe be more explicit here, for example:

```suggestion
        /// Disable native delayed delivery infrastructure. This property is a workaround allowing ServiceControl to disable delayed deliveries when setting up message pumps for error and audit queues
```

src/NServiceBus.Transport.SQS/MessagePump.cs
mauroservienti, line 1: All the new ifs in here might be addressed by having a sort of NoOpDelayedMessagePump that does nothing.

mauroservienti (approved):

johnsimons:
Raised in https://github.com/Particular/PlatformInternals/issues/659

src/NServiceBus.Transport.SQS/MessagePump.cs
johnsimons: We raised another issue to hopefully fix this properly.

src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
ramonsmits: No, I think there are many good reasons to just not have it enabled. Especially when not using delayed delivery at all or with native integration

src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
```diff
@@ -57,6 +57,11 @@ internal set
...
+        /// </summary>
+        internal bool DisableDelayedDelivery { get; set; } = false;
```
awright18, line 63: I understand this property name has been used in various places. I'm not recommending a change at this point, but I'm pointing out that using double negatives can sometimes be confusing to readers. If EnableDelayedDelivery were set to true instead, it might be easier to comprehend quickly. This is something to consider for future cases when instincts might lead you to code double negative defaults.

Issue #2520, opened 2024-07-15
`DoNotWrapOutgoingMessages` set to true cannot be used with NServiceBus.Metrics.ServiceControl


tmasternak:
#### Description

Setting [DoNotWrapOutgoingMessages](https://docs.particular.net/transports/sqs/configuration-options#do-not-wrap-message-payload-in-a-transport-envelope) to `true` causes [the performance monitoring ServiceControl plugin](https://docs.particular.net/monitoring/metrics/install-plugin) to log exceptions when sending raw data monitoring reports. 

#### Expected behavior

This configuration should work without a problem.

#### Actual behavior

Endpoints that use the SQS transport and try to send performance monitoring reports to ServiceControl, log the errors similar to:

``` code
Amazon.SQS.Model.InvalidMessageContentsException: Invalid binary character '#x1' was found in the message body, the set of allowed characters is #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF].
 ---> Amazon.Runtime.Internal.HttpErrorResponseException: Exception of type 'Amazon.Runtime.Internal.HttpErrorResponseException' was thrown.
   at Amazon.Runtime.HttpWebRequestMessage.GetResponseAsync(CancellationToken cancellationToken)
   at Amazon.Runtime.Internal.HttpHandler`1.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.Unmarshaller.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.SQS.Internal.ValidationResponseHandler.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.ErrorHandler.InvokeAsync[T](IExecutionContext executionContext)
   --- End of inner exception stack trace ---
   at Amazon.Runtime.Internal.HttpErrorResponseExceptionHandler.HandleExceptionStream(IRequestContext requestContext, IWebResponseData httpErrorResponse, HttpErrorResponseException exception, Stream responseStream)
   at Amazon.Runtime.Internal.HttpErrorResponseExceptionHandler.HandleExceptionAsync(IExecutionContext executionContext, HttpErrorResponseException exception)
   at Amazon.Runtime.Internal.ExceptionHandler`1.HandleAsync(IExecutionContext executionContext, Exception exception)
[...]

PR #2623, opened 2024-11-15
Handle sending and receiving metrics messages

mauroservienti:
Fixes:

- https://github.com/Particular/NServiceBus.AmazonSQS/issues/2520


src/NServiceBus.AmazonSQS.sln
```diff
@@ -22,6 +22,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		..\.github\workflows\release.yml = ..\.github\workflows\release.yml
 	EndProjectSection
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NServiceBus.Transport.SQS.TransportTests.DoNotWrapOutgoingMessages", "NServiceBus.Transport.SQS.TransportTests.DoNotWrapOutgoingMessages\NServiceBus.Transport.SQS.TransportTests.DoNotWrapOutgoingMessages.csproj", "{ADB39452-CFA5-4F53-9024-FCEF78A8693E}"
```
andreasohlund, line 25: Why do we need a separate project for this?

andreasohlund: I think that using the "variant" approach that SQLP does might be a better approach, happy to jump on a call and talk it through

https://github.com/Particular/NServiceBus.Persistence.Sql/blob/master/src/SqlPersistence.PersistenceTests/SqlTestVariant.cs

mauroservienti: @andreasohlund, as far as I can tell, there isn't support for variants in transport tests

andreasohlund: Talked to @mauroservienti and there might be a way to get this working, I will do a spike

lailabougria (approved):

PR #2643, opened 2024-11-27
Add all message attributes to transport operation headers for native messages

ramonsmits:
Copied from:

- https://github.com/Particular/NServiceBus.AmazonSQS/pull/2613


ramonsmits:
Not needed, `Amazon.SQS.Model.Message` is added to incoming message context to receive native message

Issue #2630, opened 2024-11-19
Amazon SQS transport doesn't support binary serialization


lailabougria:
Currently, NServiceBus supports [Json, XML and System.Text.Json serializers](https://docs.particular.net/nservicebus/serialization/) by default. However, it should also be possible to use a [custom serializer](https://docs.particular.net/nservicebus/serialization/custom-serializer).

#### Expected behavior

Plugging in a custom binary serializer should work without issues, allowing the deserialization of incoming messages and serialization of outgoing messages.

#### Actual behavior

Currently, any binary serialization mechanism would cause the transport to throw when ingesting messages, with the following exception:

`Amazon.SQS.Model.InvalidMessageContentsException: Invalid binary character '#x1' was found in the message body, the set of allowed characters is #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF].`

### Steps to reproduce

- Plug in a custom binary serializer
- Attempt to ingest messages that are serialized with a binary serializer

#### Workarounds

Use a different serialization mechanism.

#### Possible solutions

This was solved for [ServiceControl metrics messages](https://github.com/Particular/NServiceBus.AmazonSQS/issues/2520), so a similar fix may be possible. Ideally, the headers would also include the encoding type in addition to the [content type](https://docs.particular.net/nservicebus/messaging/headers#serialization-headers-nservicebus-contenttype), allowing the transport to identify the encoding type to enable correct deserialization without guessing.

mauroservienti:
It should be possible to leverage the [Base64.IsValid](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.text.base64.isvalid?view=net-9.0#system-buffers-text-base64-isvalid(system-readonlyspan((system-byte))-system-int32@)) method to check the incoming raw payload. There might be the need to backport the method implementation to older framework version, if we need to support them.

andreasohlund:
Fixed in https://github.com/Particular/NServiceBus.AmazonSQS/pull/2646

PR #2646, opened 2024-11-28
When not wrapping fallback to base64 encoding in case invalid SQS chars are being used

danielmarbach:
This is a follow-up on:

- https://github.com/Particular/NServiceBus.AmazonSQS/pull/2623

Which also addresses:

- https://github.com/Particular/NServiceBus.AmazonSQS/issues/2630

Originally, `DoNotWrapOutoingMessages` was designed to no longer base64 encode and wrap the message with the transport message. The idea was to purely hand over whatever we got to the transport. So in cases of raw dispatches when the message contains characters that are not compliant with the [SQS message encoding constraints](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html), the transport would throw.

This has consequences for scenarios like sending metrics when the payload uses raw bytes that may be incompatible with the text restrictions of SQS. By discovering that an encoded payload has characters that are not suitable to be sent and then falling back to base64 encoding, those scenarios are handled.

The transport pump already assumes when it receives non-native messages that the body might be base4 encoded and therefore gracefully falls back to reading those encoded special cases too without any extra code required.

The consequence of the change is that when not wrapping, the payload needs to be inspected for illegal characters which has computational overhead. The current implementation uses the most straightforward approach using a source-gened regex that should be reasonably fast enough. Future optimizations could be done by checking the byte array directly for invalid byte sequences / code points.


src/NServiceBus.Transport.SQS.TransportTests.DoNotWrapOutgoingMessages/Sending_messages_with_invalid_sqs_chars.cs
```diff
@@ -1,38 +1,47 @@
...
+                    return messageProcessed.SetCompleted(context);
+                },
                 (_, __) => Task.FromResult(ErrorHandleResult.Handled),
                 TransportTransactionMode.None);
 
             var headers = new Dictionary<string, string>
             {
-                { Transport.SQS.Constants.MetricsMessageMetricTypeHeaderKey, "doesn't matter" },
-                { Headers.ContentType, Transport.SQS.Constants.MetricsMessageContentTypeHeaderValue }
+                { "SomeHeader", "doesn't matter" },
```
andreasohlund, line 30: Would it make sense to add some invalid characters here as well to make this more future proof should we switch to using [message attributes](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-metadata.html) in the future when the "max 10" constraint have been lifted?

src/NServiceBus.Transport.SQS/MessageDispatcher.cs
```diff
@@ -482,6 +484,9 @@ long CalculateDelayedDeliverySeconds(UnicastTransportOperation transportOperatio
             return (body, headers);
         }
 
+        [GeneratedRegex(@"^[\u0009\u000A\u000D\u0020-\uD7FF\uE000-\uFFFD\u10000-\u10FFFF]*$", RegexOptions.Singleline)]
+        private static partial Regex ValidSqsCharacters();
```
andreasohlund, line 488:
```suggestion
       static partial Regex ValidSqsCharacters();
```

src/NServiceBus.Transport.SQS.AcceptanceTests/Sending/When_sending_messages_with_invalid_sqs_chars.cs
```diff
@@ -0,0 +1,153 @@
...
+
+                return [.. byteList];
+            }
+        }
+
+        class Receiver : EndpointConfigurationBuilder
+        {
+            public Receiver() => EndpointSetup<DefaultServer>();
+
+            public class DoTheThing : Feature
```
andreasohlund, line 119:
```suggestion
            public class CatchControlMessage : Feature
```

src/NServiceBus.Transport.SQS/MessageDispatcher.cs
```diff
@@ -17,7 +18,7 @@ namespace NServiceBus.Transport.SQS
     using Settings;
     using Transport;
 
-    class MessageDispatcher : IMessageDispatcher
+    partial class MessageDispatcher : IMessageDispatcher
```
andreasohlund, line 21: Why do we need a partial here?

andreasohlund (approved):

src/NServiceBus.Transport.SQS/MessageDispatcher.cs
danielmarbach: Partial methods with return types require a visibility modifier

src/NServiceBus.Transport.SQS/MessageDispatcher.cs
danielmarbach: The source gen for the regex requires that

lailabougria (approved):

andreasohlund:
The regex adds 12ms when used so that all seems fine

image

PR #2654, opened 2024-12-08
Set up .NET 9 to comply with global.json requirements

mauroservienti:


danielmarbach (approved):

DavidBoike:
Is it possible to add a test that will fail in CI if any of this is set up wrong? Otherwise we’ll have the same problem for every new .NET Major.

PR #2768, opened 2025-03-24
Add a transport-level configuration option for message payload byte reservation

### LINKED ISSUE #2750: DataDog Trace Propagation Header added after message size check, causes failure.
#### Is your improvement related to a problem? Please describe.
The MessageDispatcher [checks the message size](https://github.com/Particular/NServiceBus.AmazonSQS/blob/25ab81c899d273c6b4f2045576f14109be1f19d0/src/NServiceBus.Transport.SQS/MessageDispatcher.cs#L368) to determine if S3 should be used for a large message. At the point of the check, the only attributes on the SQS message (when not using message wrapping) are `NServiceBus.AmazonSQS.Headers` and `NServceBus.MessageId`. So the size is calculated before DataDog adds its own trace context propagation header, `_datadog`. This attribute isn't added until the SQS/SNS operation is actually performed by the `ISqsClient`/`ISnsClient`. This means when the message size is _just_ under the threshold for using S3, the `_datadog` attribute tips it over the edge causing the SQS/SNS operation to fail.

Unfortunately, the maximum message size is not configurable. It uses this [constant value](https://github.com/Particular/NServiceBus.AmazonSQS/blob/25ab81c899d273c6b4f2045576f14109be1f19d0/src/NServiceBus.Transport.SQS/TransportConstraints.cs#L7). If it were configurable, we could setup our endpoints to leave some room for the `_datadog` attribute.

A work around we can use for now is to disable the trace context propagation (solution found in [this related issue](https://github.com/DataDog/dd-trace-dotnet/issues/4499)) But it would be fantastic if you could make the max message size configurable.

We've actually been struggling with this issue since July! We've been trying to migrate off of SQL Transport, and this strange issue has been plaguing us! It's so sporadic, we didn't know what the cause was. We've actually gone through multiple support tickets to try to resolve it. Today was the first time I'd finally dug deep enough to discover the culprit was DataDog all this time.

We will, for now, try the workaround `DD_TRACE_AwsSqs_ENABLED=false`. But please, if you could make the max size configurable, or provide some other solution, that would be greatly appreciated!

Also... I wonder how this works with the OTel trace propagation?

#### Describe the suggested solution

Allow this [constant value](https://github.com/Particular/NServiceBus.AmazonSQS/blob/25ab81c899d273c6b4f2045576f14109be1f19d0/src/NServiceBus.Transport.SQS/TransportConstraints.cs#L7) to be configurable to let us reduce the size enough to allow room for the _datadog_ attribute.
        
#### Describe alternatives you've considered

Disabling DataDog trace propagation. This is not desirable because we loose visiblity in to the SQS and SNS operations being performed.

### Additional Context

_No response_

lailabougria:
Closes #2750

### DIFF
--- a/src/NServiceBus.Transport.SQS.Tests/ApprovalFiles/APIApprovals.ApproveSqsTransport.approved.txt
+++ b/src/NServiceBus.Transport.SQS.Tests/ApprovalFiles/APIApprovals.ApproveSqsTransport.approved.txt
@@ -80,6 +80,7 @@ namespace NServiceBus
             "ersion 8.0.0.", true)]
         public bool EnableV1CompatibilityMode { get; set; }
         public System.TimeSpan MaxTimeToLive { get; set; }
+        public long PayloadPaddingInBytes { get; set; }
         public NServiceBus.PolicySettings Policies { get; }
         public System.Func<string, string, string> QueueNameGenerator { get; set; }
         public string QueueNamePrefix { get; set; }
--- a/src/NServiceBus.Transport.SQS.Tests/ApprovalFiles/APIApprovals.ApproveSqsTransport.approved.txt
+++ b/src/NServiceBus.Transport.SQS.Tests/ApprovalFiles/APIApprovals.ApproveSqsTransport.approved.txt
@@ -80,10 +80,10 @@ namespace NServiceBus
             "ersion 8.0.0.", true)]
         public bool EnableV1CompatibilityMode { get; set; }
         public System.TimeSpan MaxTimeToLive { get; set; }
-        public long PayloadPaddingInBytes { get; set; }
         public NServiceBus.PolicySettings Policies { get; }
         public System.Func<string, string, string> QueueNameGenerator { get; set; }
         public string QueueNamePrefix { get; set; }
+        public long ReserveBytesInMessageSizeCalculation { get; set; }
         public NServiceBus.S3Settings S3 { get; set; }
         public Amazon.SimpleNotificationService.IAmazonSimpleNotificationService SnsClient { get; }
         public Amazon.SQS.IAmazonSQS SqsClient { get; }
--- a/src/NServiceBus.Transport.SQS.Tests/MessageDispatcherTests.cs
+++ b/src/NServiceBus.Transport.SQS.Tests/MessageDispatcherTests.cs
@@ -32,7 +32,7 @@
             var mockSqsClient = new MockSqsClient();
 
             var dispatcher = new MessageDispatcher(new SettingsHolder(), mockSqsClient, null, new QueueCache(mockSqsClient,
-                dest => QueueCache.GetSqsQueueName(dest, "")), null, null, 15 * 60);
+                dest => QueueCache.GetSqsQueueName(dest, "")), null, null, 15 * 60, 0);
 
             var transportOperations = new TransportOperations(
                 new TransportOperation(
@@ -76,7 +76,7 @@
             var mockSqsClient = new MockSqsClient();
 
             var dispatcher = new MessageDispatcher(new SettingsHolder(), mockSqsClient, null, new QueueCache(mockSqsClient,
-                dest => QueueCache.GetSqsQueueName(dest, "")), null, null, 15 * 60);
+                dest => QueueCache.GetSqsQueueName(dest, "")), null, null, 15 * 60, 0);
 
             var expectedId = "1234";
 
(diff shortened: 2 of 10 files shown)


mauroservienti (approved):

src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
```diff
@@ -48,6 +48,13 @@ internal void SetupSnsClient(IAmazonSimpleNotificationService snsClient, bool ex
         /// </summary>
         public string QueueNamePrefix { get; set; }
 
+        /// <summary>
+        /// Specifies an arbitrary number of bytes that will be added to the calculated payload size
+        /// which is useful to account for any overhead of message attributes determining whether
+        /// the message payload will be stored in S3 or not.
+        /// </summary>
+        public long PayloadPaddingInBytes { get; set; }
```
danielmarbach, line 56: I'm not entirely sure if PayloadPadding is a good name. What does Payload mean in this context and what verbiage would be probably closest to the ecosystem that someone could understand it. For example wearing my "I have some understanding of the various ways the transport treats the payload" I can see how there are multiple ways with wrapping, not wrapping and native integration scenarios how the payload is treated and even when I use that phrasing I'm not sure whether it is clear that I mean the business data when I elaborate on the payload or whether the term payload represents the message payload which can be the business payload or the transport message. 

Given that SQS itself uses the `MaximumMessageSize` I wonder if a more "bare metal"  language like `MaximumMessageSizePaddingInBytes` would be simpler to understand. Especially considering that we might add the MaximumMessageSizeInBytes setting in other iterations, I like that but I can see how arguments can be made both ways.

src/NServiceBus.Transport.SQS/SnsPreparedMessage.cs
```diff
@@ -21,13 +21,15 @@ public string MessageId
         public string Body { get; set; }
         public string Destination { get; set; }
         public long Size { get; private set; }
+        public long PayloadPaddingInBytes { get; set; }
```
danielmarbach, line 24: It might be a departure of how we already treat this type but given this property is kind of mandatory I wonder if we should make this init only and have further iterations to also clean up the other properties in a dedicated PR.

src/NServiceBus.Transport.SQS.Tests/MessageDispatcherTests.cs
```diff
@@ -926,8 +926,42 @@ public async Task Should_handle_empty_body_gracefully(bool wrapMessage)
             Assert.That(request.MessageBody, Is.Not.Null.Or.Empty);
         }
 
-        interface IEvent { }
+        [Test]
+        public async Task Does_bump_to_S3_when_payload_padding_is_set()
```
danielmarbach, line 930: I haven't checked all the tests for a consistent verbiage, but on a superficial level, it looks like we use the term "upload" instead of bump.

src/NServiceBus.Transport.SQS.Tests/ApprovalFiles/APIApprovals.ApproveSqsTransport.approved.txt
```diff
@@ -80,6 +80,7 @@ namespace NServiceBus
             "ersion 8.0.0.", true)]
         public bool EnableV1CompatibilityMode { get; set; }
         public System.TimeSpan MaxTimeToLive { get; set; }
+        public long PayloadPaddingInBytes { get; set; }
```
danielmarbach, line 83: See comment below

src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
lailabougria: I don't like the addition of `MaximumMessageSize`, especially at the beginning, which doesn't immediately convey the intent of the padding. The name then seems to convey that the message size is increased, which it isn't. 

Here are some other options, but they are very very long:
- `AddPaddingInBytesToMessageSizeCalculation`
- `IncludePaddingInBytesToMessageSizeCalculation`

What do you think?

danielmarbach: How about we just call it `MessageSizePaddingInBytes` then?

lailabougria: I like @mauroservienti's suggestion, which also gets rid of the "padding" terminology and expresses the intent better.

PR #2773, opened 2025-03-27
Add Support for Automatic SQS Message Visibility Timeout Renewal

danielmarbach:
This PR introduces automatic message visibility renewal for the SQS transport. This mechanism ensures messages are not prematurely reprocessed if the handler execution time exceeds the visibility timeout configured on the queue or the receive request.

Without visibility renewal, long-running message handlers may experience:
- Duplicate processing due to expired visibility
- Message deletion failures, resulting in message reappearance
- Misleading delivery attempt counts for poison message detection

## Settings

MaxAutoMessageVisibilityRenewalDuration (default: 5 minutes)

Controls how long visibility timeout renewals will be attempted while processing a message.

```csharp
transport.MaxAutoMessageVisibilityRenewalDuration = TimeSpan.FromMinutes(10);
```

- Renewal is disabled if set to `TimeSpan.Zero
`- Max value: 12 hours (SQS maximum visibility timeout)

MessageVisibilityTimeout (default: null)

Overrides the default queue visibility timeout used when receiving messages.

- If unset, the queue’s visibility timeout is retrieved using `GetQueueAttributes`
- If set, it affects:
  - The visibility of messages upon receive
  - The renewal calculation (renew after = ½ remaining - buffer)

Both properties include runtime validation to ensure values fall within SQS’s supported range of 0 < timeout ≤ 12h.

## Renewal flow

```mermaid
flowchart TD
  Start("Start Renewal") --> Wait[/"Wait before renewal (dynamic buffer)"/]
  Wait --> CheckCancelled{"Renewal Token Cancelled?"}
  CheckCancelled -- Yes --> Done["Stop renewal, log"]
  CheckCancelled -- No --> Extend["Extend visibility"]
  Extend --> UpdateTime["Update next expiry"]
  UpdateTime --> Wait
```

```mermaid
flowchart TD
  A["Start processing"] --> B["Start visibility renewal loop"]
  B --> C["Linked cancellation token created"]
  C --> D{"Processing finished or renewal stopped?"}
  D -- Yes --> E["Cancel renewal, exit"]
  D -- No --> F{"Renewal failure (e.g., timeout, receipt invalid)?"}
  F -- Yes --> G["Cancel processing"]
  F -- No --> H["Continue processing"]
  H --> D
```

### Edge case handling

|   Scenario                                                    |   Outcome                                                                    |
|---------------------------------------------------------------|------------------------------------------------------------------------------|
|   Visibility expires before renewal                           |   Renewal will extend it retroactively if the receipt handle is still valid  |
|   Renewal fails due to ReceiptHandleIsInvalidException        |   The message is considered lost, processing is canceled                     |
|   Handler finishes early                                      |   Renewal is cleanly canceled                                                |
|   Renewal fails due to message picked up by another consumer  |   Graceful cancellation of processing logic                                  |
|   Message deletion fails post-processing                      |   Marked in-memory for future deletion attempts                              |


src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -249,6 +272,9 @@ async Task ProcessMessageSwallowExceptionsAndReleaseMaxConcurrencySemaphore(Mess
             }
             finally
             {
+                await coordinationTokenSource.CancelAsync()
+                    .ConfigureAwait(false);
```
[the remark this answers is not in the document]
danielmarbach: This is incorrect

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -220,13 +210,46 @@ async Task PumpMessagesAndSwallowExceptions(CancellationToken messagePumpCancell
...
+                    await Task.Delay(visibilityTimeout / 2, cancellationToken)
+                        .ConfigureAwait(false);
+                    // we don't want this to be cancellable because we are doing best-effort to complete inflight messages
+                    // on shutdown.
+                    await sqsClient.ChangeMessageVisibilityAsync(
+                        new ChangeMessageVisibilityRequest
+                        {
+                            QueueUrl = inputQueueUrl,
+                            ReceiptHandle = receivedMessage.ReceiptHandle,
+                            VisibilityTimeout = 15 + 30
```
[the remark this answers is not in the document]
danielmarbach: This code wouldn't even compile

src/NServiceBus.Transport.SQS/Renewal.cs
```diff
@@ -0,0 +1,163 @@
...
+using System.Threading.Tasks;
+using Amazon.SQS;
+using Amazon.SQS.Model;
+using Logging;
+
+static class Renewal
+{
+    // This method does not check whether the visibility time has already expired. The reason being that it is possible to renew the visibility
+    // even when the visibility time has expired as long as the message has not been picked up by another consumer or receive attempt the
+    // original receipt handle is still valid.
```
mauroservienti, line 14:
```suggestion
    // This method does not check whether the visibility time has expired. That is because it's possible to renew the visibility even when the visibility time has expired.
    // The receipt handle remains valid until another consumer or a receiving attempt picks up the message.
```

src/NServiceBus.Transport.SQS/Renewal.cs
```diff
@@ -0,0 +1,163 @@
...
+    public static TimeSpan CalculateRenewalTime(DateTimeOffset visibilityTimeExpiresOn,
+        TimeProvider timeProvider = null)
+    {
+        timeProvider ??= TimeProvider.System;
+        var remainingTime = visibilityTimeExpiresOn - timeProvider.GetUtcNow();
+
+        if (remainingTime < TimeSpan.FromMilliseconds(400))
+        {
+            return TimeSpan.Zero;
+        }
```
mauroservienti, line 152: Shall we document what's unique about 400ms?

mauroservienti: I also guess that `return TimeSpan.Zero;` means renew immediately, right?

src/NServiceBus.Transport.SQS/Renewal.cs
```diff
@@ -0,0 +1,163 @@
...
+                }
+
+                var utcNow = timeProvider.GetUtcNow();
+                var remainingTime = visibilityExpiresOn - utcNow;
+                // It is possible that the remaining time is negative when for example the continuation task scheduling takes too long
+                // In those cases we attempt to overextend the visibility timeout by the positive remaining time in seconds
+                var calculatedVisibilityTimeout =
+                    Math.Max(Math.Abs((int)remainingTime.TotalSeconds) + visibilityTimeoutInSeconds,
+                        visibilityTimeoutInSeconds);
+                // immediately calculating the new expiry before calling updating the visibility to be on the safe side
```
mauroservienti, line 57:
```suggestion
                // immediately calculating the new expiry before updating the visibility to be on the safe side
```

src/NServiceBus.Transport.SQS/DelayedMessagesPump.cs
```diff
@@ -413,6 +413,10 @@ async Task DeleteMessage(SqsReceivedDelayedMessage messageToDeleteWithAnotherAtt
                 await sqsClient.DeleteMessageAsync(messageToDeleteWithAnotherAttempt.QueueUrl, messageToDeleteWithAnotherAttempt.ReceiptHandle, cancellationToken).ConfigureAwait(false);
             }
             catch (ReceiptHandleIsInvalidException ex)
+            {
+                Logger.Info($"Message receipt handle '{messageToDeleteWithAnotherAttempt.ReceiptHandle}' is invalid.", ex);
+            }
+            catch (AmazonSQSException ex) when (ex.IsCausedByMessageVisibilityExpiry())
```
mauroservienti, line 419: Shall we link here to the issue about the nuances of these exceptions? This could help our future self remember why two different exceptions are logging the same message.

src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
```diff
@@ -142,6 +142,53 @@ public Func<Type, string, string> TopicNameGenerator
...
+        public TimeSpan? MessageVisibilityTimeout
+        {
+            get => messageVisibilityTimeout;
+            set
+            {
+                messageVisibilityTimeout = value;
+                if (!value.HasValue)
+                {
+                    return;
+                }
```
mauroservienti, line 178: If `value` is null, `messageVisibilityTimeout` will be null. So far, so good. Why the `if` is not setting to null also the `messageVisibilityTimeoutInSeconds`?

```suggestion
                messageVisibilityTimeout = value;
                if (!value.HasValue)
                {
                    messageVisibilityTimeoutInSeconds = null;
                    return;
                }
```

src/NServiceBus.Transport.SQS/Configure/SqsTransport.cs
```diff
@@ -142,6 +142,53 @@ public Func<Type, string, string> TopicNameGenerator
...
+        public TimeSpan MaxAutoMessageVisibilityRenewalDuration
+        {
+            get => maxAutoMessageVisibilityRenewalDuration;
+            set
+            {
+                var visibilityTimeoutInSeconds = (int)value.TotalSeconds;
+                ArgumentOutOfRangeException.ThrowIfNegative(visibilityTimeoutInSeconds, nameof(MaxAutoMessageVisibilityRenewalDuration));
+                ArgumentOutOfRangeException.ThrowIfGreaterThan(visibilityTimeoutInSeconds, TimeSpan.FromHours(12).TotalSeconds, nameof(MaxAutoMessageVisibilityRenewalDuration));
+
+                maxAutoMessageVisibilityRenewalDuration = value;
```
mauroservienti, line 159: Minor improvement to avoid confusion

```suggestion
                var maxAutoMessageVisibilityTimeoutInSeconds = (int)value.TotalSeconds;
                ArgumentOutOfRangeException.ThrowIfNegative(maxAutoMessageVisibilityTimeoutInSeconds, nameof(MaxAutoMessageVisibilityRenewalDuration));
                ArgumentOutOfRangeException.ThrowIfGreaterThan(maxAutoMessageVisibilityTimeoutInSeconds, TimeSpan.FromHours(12).TotalSeconds, nameof(MaxAutoMessageVisibilityRenewalDuration));

                maxAutoMessageVisibilityRenewalDuration = value;
```

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -220,45 +219,64 @@ async Task PumpMessagesAndSwallowExceptions(CancellationToken messagePumpCancell
...
+                sqsClient, inputQueueUrl, messageVisibilityLostCancellationTokenSource, timeProvider,
+                cancellationToken: renewalCancellationTokenSource.Token);
+
+            ExceptionDispatchInfo processMessageFailed = null;
+
+            try
+            {
+                // deliberately not passing renewalCancellationTokenSource into the processing method. There are scenarios where it is possible that you can still
+                // successfully process the message even when the message visibility has expired. But when the messageVisibilityLostCancellationTokenSource gets cancelled
+                // we now that we have lost the visibility and we should not process the message anymore. This gives the users a chance to stop processing.
```
mauroservienti, line 248:
```suggestion
                // we know that we have lost the visibility and we should not process the message anymore. This gives the users a chance to stop processing.
```

src/NServiceBus.Transport.SQS.Tests/InputQueuePumpTests.cs
```diff
@@ -259,6 +267,119 @@ await SetupInitializedPump(onMessage: (ctx, ct) =>
...
+
+            // The message expires in 30 seconds
+            var visibilityExpiresOn = timeProvider.Start.AddSeconds(30);
+
+            var task = pump.ProcessMessageWithVisibilityRenewal(message, visibilityExpiresOn, timeProvider, CancellationToken.None);
+
+            // Simulate the time passing. Default visibility timeout is 30 seconds.
+            timeProvider.Advance(TimeSpan.FromSeconds(16));
+            timeProvider.Advance(TimeSpan.FromSeconds(32));
+            timeProvider.Advance(TimeSpan.FromSeconds(60));
```
mauroservienti, line 291: Purely for my OCD 😬

```suggestion
            timeProvider.Advance(TimeSpan.FromSeconds(64));
```

src/NServiceBus.Transport.SQS/DelayedMessagesPump.cs
lailabougria: Isn't most of that covered in the PR description?

src/NServiceBus.Transport.SQS/Renewal.cs
lailabougria: Yes, that's used then in Task.Delay, which then won't delay basically

src/NServiceBus.Transport.SQS/Renewal.cs
```diff
@@ -0,0 +1,163 @@
...
+        if (Logger.IsDebugEnabled)
+        {
+            Logger.DebugFormat("The message visibility timeout renewal for message with native ID '{0}' was stopped. The message visibility might expire at '{1}' UTC", receivedMessage.MessageId, visibilityExpiresOn);
+        }
+    }
+
+    public enum Result
+    {
+        Failed,
+        Stopped
```
lailabougria, line 140: Personally I prefer Canceled here.

danielmarbach: The message visibility renewal algorithm has two internal outcomes. It is either failed or canceled (implementation detail) . I modeled the result (external) as failed and stopped because I felt the fact that the underlying renewal is canceled with cancellation tokens doesn't necessarily warrant calling it canceled at the result level. To me, canceled still carries a connotation of something being aborted or cut short unexpectedly, whereas in this case, the stop is often intentional or benign—like when the message is processed and there’s no need to extend visibility anymore.

I also felt that having failed and canceled as the two outcomes makes it sound like both represent errors or undesired states, while stopped feels more natural as a controlled, expected end. It creates a clearer mental distinction between “something went wrong” and “we’re done here.”

 I’m still on the fence—mainly because I think stopped communicates better that the renewal simply didn’t need to continue, rather than implying a problem occurred.

src/NServiceBus.Transport.SQS/DelayedMessagesPump.cs
danielmarbach: Plus they are not really logging the same message ;)

PR #2774, opened 2025-03-27
Propagate native headers to NServiceBus headers

mauroservienti:
This PR propagates incoming native message attributes to NServiceBus headers. If the incoming message attribute key matches an NServiceBus header key, the NServiceBus header value will be overwritten by the native message attribute value with a matching key.

The following transport-specific headers are already present on the incoming native message as message attributes:

- `NServiceBus.AmazonSQS.TimeToBeReceived`
- `NServiceBus.AmazonSQS.DelaySeconds`
- `NServiceBus.AmazonSQS.Headers`
- `NServiceBus.AmazonSQS.S3BodyKey`
- `NServiceBus.AmazonSQS.MessageTypeFullName`

They are not promoted to NServiceBus headers as the internal transport implementation uses them to determine how to handle incoming messages.

## PoA

- [x] Add PR description and labels
- [x] Documentation — https://github.com/Particular/docs.particular.net/pull/7066

### DIFF
--- a/src/NServiceBus.Transport.SQS/InputQueuePump.cs
+++ b/src/NServiceBus.Transport.SQS/InputQueuePump.cs
@@ -442,6 +442,7 @@ namespace NServiceBus.Transport.SQS
         {
             foreach (var receivedMessageMessageAttribute in receivedMessage.MessageAttributes)
             {
+                // TODO is there a better way to achieve the same result? This seems fragile
                 if (TransportHeaders.AllTransportHeaders.Contains(receivedMessageMessageAttribute.Key) || receivedMessageMessageAttribute.Key == Headers.MessageId)
                 {
                     continue;
--- a/src/NServiceBus.Transport.SQS/InputQueuePump.cs
+++ b/src/NServiceBus.Transport.SQS/InputQueuePump.cs
@@ -433,7 +433,21 @@ namespace NServiceBus.Transport.SQS
             }
             // HINT: Message Id is the only required header
             transportMessage.Headers.TryAdd(Headers.MessageId, messageIdOverride);
+            AddCustomNativeHeadersToNServiceBusHeaders(receivedMessage, transportMessage);
+
             return transportMessage;
+        }
+
+        static void AddCustomNativeHeadersToNServiceBusHeaders(Message receivedMessage, TransportMessage transportMessage)
+        {
+            foreach (var receivedMessageMessageAttribute in receivedMessage.MessageAttributes)
+            {
+                if (TransportHeaders.AllTransportHeaders.Contains(receivedMessageMessageAttribute.Key) || receivedMessageMessageAttribute.Key == Headers.MessageId)
+                {
+                    continue;
+                }
+                _ = transportMessage.Headers.TryAdd(receivedMessageMessageAttribute.Key, receivedMessageMessageAttribute.Value?.StringValue ?? string.Empty);
+            }
         }
 
         static bool CouldBeNativeMessage(TransportMessage msg)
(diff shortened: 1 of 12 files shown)


src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -415,9 +415,24 @@ public static TransportMessage ExtractTransportMessage(Message receivedMessage,
...
+        static void AddCustomNativeHeadersToNServiceBusHeaders(Message receivedMessage, TransportMessage transportMessage)
+        {
+            foreach (var receivedMessageMessageAttribute in receivedMessage.MessageAttributes)
+            {
+                // TODO is there a better way to achieve the same result? This seems fragile
+                if (TransportHeaders.AllTransportHeaders.Contains(receivedMessageMessageAttribute.Key) || receivedMessageMessageAttribute.Key == Headers.MessageId)
+                {
+                    continue;
+                }
+                _ = transportMessage.Headers.TryAdd(receivedMessageMessageAttribute.Key, receivedMessageMessageAttribute.Value?.StringValue ?? string.Empty);
```
mauroservienti, line 432: `TryAdd` makes it so that NServiveBus headers will never be overridden. Should native headers be able to override NServiceBus headers?

danielmarbach: We have discussed this and concluded for native integration scenarios it makes sense for the message attributes to take precedence when there are duplicates. The likelihood of duplicates is very low anyway.

src/NServiceBus.Transport.SQS/InputQueuePump.cs
```diff
@@ -415,9 +415,24 @@ public static TransportMessage ExtractTransportMessage(Message receivedMessage,
...
+            AddCustomNativeHeadersToNServiceBusHeaders(receivedMessage, transportMessage);
+
             return transportMessage;
         }
 
+        static void AddCustomNativeHeadersToNServiceBusHeaders(Message receivedMessage, TransportMessage transportMessage)
+        {
+            foreach (var receivedMessageMessageAttribute in receivedMessage.MessageAttributes)
+            {
+                // TODO is there a better way to achieve the same result? This seems fragile
```
lailabougria, line 427: You can probably filter the ones out you don't want with a linq query, which will improve readability too, but possibly it's slower (that would be Daniel's department ;-))

mauroservienti: @danielmarbach @lailabougria we need to discuss the `if` in this `foreach`. It's super fragile.

danielmarbach: Like I mentioned in slack, I think we can override all matching ones with the native. That would be consistent with the description we have in the doc following the principle of native headers win

danielmarbach:
There is something I can't entirely figure out why it bothers me so please bear with me. It seems for the prepared messages with have some kind of helper methods that allow to "Copy" native headers or "remove" native headers while we don't have the same for the transport message. It also seems that we have existing code that deliberately copies some headers (like the MessageTypeFullName) while the later code then excludes them again which I think increases the cognitive burden the code has. I wonder if it would be better to switch to a model where we encapsulate that in the transport message somehow and then cleanup the rest of the code base or make it a very explicit model where we copy over all by default and then remove the ones that are not needed but in all other code paths no longer explicitly add things. On the fence what is the best approach to be honest...

mauroservienti:
I cannot agree more, @danielmarbach. The incoming message-to-transport message handling is a mess due to all the corner cases we have accumulated over the years. To support that, it's worth noting that as soon as I removed the message id guard (https://github.com/Particular/NServiceBus.AmazonSQS/pull/2774/commits/0910136272f475d7a507befc48a4df28775f1c25) a long list of tests started failing, proving that it's a bit of a chaos.

lailabougria (approved):

PR #2777, opened 2025-03-28
Make sure sdk clients are disposed even if shutdown is cancelled

andreasohlund:


mauroservienti (approved):

danielmarbach:
What's the point of implementing disposable when the code that creates the infrastructure doesn't use it?

andreasohlund:
The idea was to have core start calling dispose in the next minor/major?

danielmarbach:
Agreed but all that is not described in this PR. I think it would be great to describe this in the PR and hint that we are trying to avoid doing multiple releases. Then we need to also check whether the code handles double disposal properly because the clients would be disposed multiple times (the clients should if they follow the pattern properly)

andreasohlund:
After discussing with @danielmarbach we came to the conclusion that implementing IDisposble is premature and should be driven from the core transport seam requiring it rather than the other way around.

The following issues have been raised to track that future work

- https://github.com/Particular/NServiceBus/issues/7322
- https://github.com/Particular/NServiceBus/issues/7323

danielmarbach (approved):

ramonsmits:
> What's the point of implementing disposable when the code that creates the infrastructure doesn't use it?

That isn't entirely true @danielmarbach as transport infrastructure can be used independently from core like done by the connector, raw messaging and the bridge.

danielmarbach:
@ramonsmits The discussion I had with @andreasohlund was much more nuanced because there are larger consequences of implementing IDisposable when you follow that pattern.

PR #2788, opened 2025-04-08
File scoped namespace

danielmarbach:


mauroservienti (approved):
The GitHub UI freezes every now and then 🤷

PR #2790, opened 2025-04-08
Structuring

danielmarbach:


src/NServiceBus.Transport.SQS/Administration/QueueCreator.cs
danielmarbach, line 1: On the fence about having a single file in a folder, but some other transports use a similar structure, so I went with it for now. Happy to put it back into the root

mauroservienti (approved):

PR #2850, opened 2025-07-17
Swap Dependabot for Renovate to auto approve AWS dependencies

tamararivera:


DavidBoike:
We have the additional requirement of an approving review, so Renovate can't auto-merge all by itself. So we could use Renovate to set automerge and something else (this workflow, InternalAutomation) to approve, but for now this keeps the automerge capability encapsulated in a single file. We might re-evaluate next week.

DavidBoike (approved):
We'll see how it works in this repo and then decide what to do about others that reference AWSSDK packages.

PR #3107, opened 2026-06-08
Mark message-driven pub/sub compatibility mode as obsolete

mnadareski-particular:
All uses of `EnableMessageDrivenPubSubCompatibilityMode` need to be marked as obsolete with a warning starting in the next minor version.

This also adds the `#pragma warning disable/restore` directives for tests only.

### DIFF
--- a/src/NServiceBus.Transport.SQS.AcceptanceTests/NativePubSub/HybridModeRateLimit/When_publishing_one_event_type_to_native_and_non_native_subscribers_in_a_loop.cs
+++ b/src/NServiceBus.Transport.SQS.AcceptanceTests/NativePubSub/HybridModeRateLimit/When_publishing_one_event_type_to_native_and_non_native_subscribers_in_a_loop.cs
@@ -68,7 +68,9 @@ public class When_publishing_one_event_type_to_native_and_non_native_subscribers
             {
                 b.CustomConfig(config =>
                 {
+#pragma warning disable CS0618 // Type or member is obsolete
                     var migrationMode = config.ConfigureRouting().EnableMessageDrivenPubSubCompatibilityMode();
+#pragma warning restore CS0618 // Type or member is obsolete
                     migrationMode.SubscriptionsCacheTTL(testCase.SubscriptionsCacheTTL);
                     migrationMode.TopicCacheTTL(testCase.NotFoundTopicsCacheTTL);
                 });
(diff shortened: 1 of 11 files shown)


mauroservienti (approved):

src/NServiceBus.Transport.SQS/Configure/SqsTransportSettings.cs
```diff
@@ -291,9 +291,10 @@ public static class MessageDrivenPubSubCompatibilityModeConfiguration
     /// Enables compatibility with endpoints running on message-driven pub-sub
     /// </summary>
     /// <param name="transportExtensions">The transport to enable pub-sub compatibility on</param>
-    [PreObsolete("https://github.com/Particular/NServiceBus/issues/6471",
-        Note = "Hybrid pub/sub support cannot be obsolete until there is a viable migration path to native pub/sub",
-        Message = "Hybrid pub/sub is no longer supported, use native pub/sub instead")]
+    [Obsolete("Hybrid pub/sub is no longer supported, use native pub/sub instead. Will be treated as an error from version 10.0.0. Will be removed in version 11.0.0.", false)]
```
bording: While this was the message in the `PreObsolete`, I don't think this is quite right. It's just a warning, and fully works, so it's not yet "no longer supported."

src/NServiceBus.Transport.SQS.AcceptanceTests/NativePubSub/HybridModeRateLimit/When_publishing_one_event_type_to_native_and_non_native_subscribers_in_a_loop.cs
```diff
@@ -68,7 +68,9 @@ public async Task Should_not_rate_exceed(TestCase testCase)
             {
                 b.CustomConfig(config =>
                 {
+#pragma warning disable CS0618 // Type or member is obsolete
```
bording, line 71: For all of these pragmas, if we can't go ahead and remove the obsolete code now, then we need some sort of instructions as comments to explain what actions need to be taken when the warnings become errors.

For example, can we just delete the hybrid mode call and leave the test, or should the entire test be deleted?

That's the sort of thing I'd want to see if I was the person having to clean these up in the future.

